PR #4, opened 2017-12-24
npm3 uses a different cache structure

adityashah1212:
This pr is to support npm3

### DIFF
--- a/npm/flatpak-npm-generator.py
+++ b/npm/flatpak-npm-generator.py
@@ -15,7 +15,7 @@ electron_arches = {
 }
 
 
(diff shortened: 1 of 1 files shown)


nedrichards:
If it's npm3 you're targeting a package.json right, rather than a package-lock.json? If so - we should probably update the documentation a bit more here since just going --npm3 won't always work will it? Or is there a different use case you're thinking of.

adityashah1212:
No, it is still using package-lock.json. This thing that it is targeting for the the difference the cache directory structure has in npm3 and npm5.

PR #5, opened 2017-12-30
generate sources from yarn.lock

adityashah1212:


nedrichards:
I am very excited to see this! Will take a closer look later.

nedrichards:
This by and large looks 'fine' from a read. I'm going to try and use it to package a simple project and if it works, happy to merge since it's an entirely new system.

I'd love to see a README.md added to give people a bit more documentation but that can be added later if you don't have time now.

adityashah1212:
there is still one thing missing in this. It seems that there are instances where git urls are given as git+https or git+http or git+ssh. I have seen it, but I am waiting on this pull request to merge. I want to keep it separate.

nedrichards:
Makes sense. I wasn't able to test this fully as the system with a new enough flatpak and flatpak-builder didn't have enough RAM to build nodejs - but I was able to test both parts enough to know that it does what it's supposed to!

PR #12, opened 2018-01-25
pip: Fix wheel names

ptomato:
The naming convention for wheel packages is given by PEP 427. It involves
more hyphens than just the one which was customary for tar.gz packages.


TingPing:
A `whl` file should never be downloaded, we pass `'--no-binary', ':all:',`.

ptomato:
Oh, that must be because I removed the `--no-binary :all:` temporarily. Never mind this PR then!

In that case, I'll open a different issue for you... the source package of lxml won't install, which is why I needed the wheel :stuck_out_tongue:

PR #23, opened 2018-05-20
Add Rubygems module generator script for Ruby applications

yuntan:
This script may help people who want to distribute Ruby applications as Flatpak.


TingPing:
Thanks!

PR #25, opened 2018-06-24
Add a basic json to yaml converter

lazka:
Nothing fancy, but has some nice defaults like preserving order and
indenting lists.

### DIFF
--- a/flatpak-json2yaml.py
+++ b/flatpak-json2yaml.py
@@ -21,15 +21,39 @@
 # SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 import os
+import re
+import sys
 import argparse
 import json
 import yaml
 from collections import OrderedDict
 
 
+def json_remove_comments(json_data):
+
+    def repl_func(match):
+        comment = match.group(1).decode("utf-8")
+        line = match.string[:match.start()].count(b"\n") + 1
+        print("Removed comment in line {}: {}".format(line, repr(comment)),
+              file=sys.stderr)
+        return b""
+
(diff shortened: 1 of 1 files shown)


flatpak-json2yaml.py
```diff
@@ -0,0 +1,73 @@
...
+    parser.add_argument('json_file', type=str,
+                        help='The flatpak JSON file to convert')
+    parser.add_argument('-o', '--output', type=str, dest='out_file',
+                        help='The yaml target path')
+    args = parser.parse_args()
+
+    with open(args.json_file, "rb") as h:
+        out_file = args.out_file
+        if out_file is None:
+            out_file = os.path.splitext(args.json_file)[0] + '.yaml'
```
TingPing: We've been using the `yml` file extension mostly. Might as well be consistent.

TingPing:
A lot of flatpak json files use comments which are non-standard, it would be nice to not fail on those.

flatpak-json2yaml.py
lazka: ok

lazka:
> A lot of flatpak json files use comments which are non-standard, it would be nice to not fail on those.

ok, I'll have a look

lazka:
https://github.com/flathub/io.github.gillesdegottex.FMIT/blob/master/io.github.gillesdegottex.FMIT.json

```
./flatpak-json2yaml.py io.github.gillesdegottex.FMIT.json
Removed comment in line 14: '/* acs_alsa support */'
Removed comment in line 16: '/* acs_jack support */'
Removed comment in line 83: '/* acs_qt does not work on Linux */'
Removed comment in line 88: '/* must be done before calling qmake, or translations will not be installed (?) */'
```

Not sure if this is enough.. ?

PR #31, opened 2018-08-05
Replace list(set()) trick with actual remove_duplicates function

SolarLiner:
Sets only allow hashable objects, allowing for a more efficient
comparison and management of itself. Dict objects aren't
hashable, therefore can't be turned into set items. We replace that
trick with an actual function that reconstructs the list but only adds
items not already present. This function uses "obj not in list" notation
which uses `__eq__` and not `__hash__` interface, which `dict`
implements.

Included is a the resulting `generated-sources`.py which show the
resulting generated sources.


generated-sources.json
```diff
@@ -0,0 +1,1109 @@
+[
```
TingPing: Extraneous file.

SolarLiner: Hm, it was in the npm folder, I thought it was also in the yarn folder as well. Should I remove it?

TingPing: Well this is in the root folder?

SolarLiner: Uh, this is embarrassing. It *was* on the root folder, and overwriting it with the actual yarn generate sources shows there is no difference, which I guess is a good regression test.

PR #35, opened 2018-09-19
add flatpak-pip-generator-version-choices

yucefsourani:


TingPing:
Just add an option to the existing script, don't copy the entire thing.

yucefsourani:
Have you tried it?
This script also provides the version selection, I did not want to modify the script already in place.

PR #51, opened 2019-03-05
Add Builder Tool for Python Poetry

danyeaw:
This new builder tool automatically creates a manifest based on the poetry.lock. Poetry has already done advanced dependency resolution, so this helps ensure that the versions of the packages used to build the Flatpak are consistent and repeatable to what Poetry is using.


danyeaw:
@TingPing Thanks ♥️👍

PR #55, opened 2019-05-18
Add flatpak-node-generator

refi64:
While working on a Scratch Flatpak, several problems with the current flatpak-npm-generator and flatpak-yarn-generator became readily apparent (I'll explain more below). So to work around this, I created flatpak-node-generator, which combines support for both npm and yarn in a much cleaner and more powerful generator script.

I haven't done any documentation or examples yet, that should be coming soon. I just figured I'd submit this here for early review.

## Advantages

flatpak-node-generator has many features and advantages over the other two scripts:

- Supports newer Node and npm versions. At some point during Node 10's release cycle, npm started *caching the registry http requests*, which is what started the initial rabbit hole into creating this new script. flatpak-npm-generator did not handle this. It also generates a cache index like npm now wants.
- Works on newer Electron versions, flatpak-npm-generator relied on filenames that have changed around three years ago (!).
- Shared code, e.g. when I added support for Chromedriver, it immediately was applied to both npm and yarn generation.
- Related to the above, first-class support for modern Electron packaging workflows (e.g. via webpack).
- Due to the first point (needing to download each http requests), a linear, one by one download in the script generation would be brutally slow. Therefore, the new script is built on Python's asyncio system, which gives it pretty nice performance given it has to make a few thousand requests during a single script generation.
- Support for multiple package-lock.json and yarn.lock was broken with Git sources in the original npm-generator, this fixes that.
- The original regex replacements on Git sources would break if there were two on different commits, this also fixes that.
- The yarn parser is indentation-based, which means that it's overall more solid and won't break if a package name contains the word "dependencies".
- Automatically retries failed requests.
- Nicer architecture, e.g. when Yarn 2 comes out it will be easy to add a new lockfile parser while retaining the same offline download logic.
- This one is largely type-safe thanks to Python's new type annotations. 

## Disadvantages

- **aiohttp is required to get better performance.** flatpak-node-generator will fall back onto urllib.request if aiohttp is not found, but all requests will be linear, meaning the performance will be terrible.
- Python 3.6+ is required. It has been out for quite a while already, and based on current plans it will probably be EOL'd in a little over a year.
- There is no support for Electron 1 or Node 8. Node 8 is reaching EOL in just a few months, so this was pretty intentional. Anyone still on Node 8 can use flatpak-npm-generator; it will work since it was created before the new npm caching nightmare started.

For an example of why aiohttp is necessary for performance, generating the sources for Scratch's Flatpak took 1 minute and 17 seconds with aiohttp. Without it, it takes over *12 minutes*.

### DIFF
--- a/node/.gitignore
+++ b/node/.gitignore
@@ -1,0 +1,1 @@
+.mypy_cache/
--- a/node/.gitignore
+++ b/node/.gitignore
@@ -1,1 +1,3 @@
 .mypy_cache/
+.flatpak-builder/
+_build/
--- a/node/README.md
+++ b/node/README.md
@@ -1,3 +1,130 @@
 # flatpak-node-generator
 
 A more modern successor for flatpak-npm-generator and flatpak-yarn-generator.
+
+## Requirements
+
+- Python 3.6+.
+- aiohttp. (flatpak-node-generator will fall back onto urllib.request if aiohttp is not available,
+  but performance will take a serious hit.)
+
+## Complete examples
+
+There are two examples provided for how to use flatpak-node-generator:
+
+- `vanilla-quick-start` - A Flatpak of
+  [electron-quick-start](https://github.com/electron/electron-quick-start). It uses npm for
+  package management and a rather basic Electron workflow.
+  (Current on the Electron 4 version.)
+- `webpack-quick-start` - A Flatpak of
+  [electron-webpack-quick-start](https://github.com/electron-userland/electron-webpack-quick-start).
+  It uses yarn for package management and electron-builder + webpack.
+
+Both manifests have comments to highlight their differences, so you can mix and match to e.g.
+get npm with electron-builder.
+
+## Usage
+
+```
+usage: flatpak-node-generator.py [-h] [-o OUTPUT] [-r] [-R RECURSIVE_PATTERN]
+                                 [--no-devel] [--no-index] [--no-aiohttp]
+                                 [--retries RETRIES] [-P]
+                                 {npm,yarn} lockfile
+
+Flatpak Node generator
+
+positional arguments:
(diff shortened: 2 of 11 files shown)


refi64:
Fixed the unused imports and except blocks, the incomplete URL doesn't really apply, and especially the use of the return value is incorrect (since it actually returns an asyncio future that's run).

refi64:
@barthalion Ah true, I also want to add some more modern examples.

x80486:
@refi64, this is great!

By the way, would you mind putting some guidelines on how to use this one? I'm currently trying to Flatpak [org.standardnotes.standardnotes](https://github.com/flathub/flathub/pull/988/files/) and it's failing left and right, so I guess I would give it a try with this one and see if I get a different result.

Thanks in advance! :sunglasses:

refi64:
Yes that will all come very, very shortly :grin:

refi64:
Okay so I just updated to fix some stuff and add some examples & documentation, @x80486 want to take a look now?

Note that the examples do use the [currently unmerged Node 10 SDK extension](https://github.com/flathub/flathub/pull/1004), which is mostly just waiting on the [fd.o sdk Python 2 fixes](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/merge_requests/1318). Other than that, please point out any issues in the examples or documentation.

x80486:
> Okay so I just updated to fix some stuff and add some examples & documentation, @x80486 want to take a look now?

I read through it and it looks great...plus the examples makes it easier for a quick-start. Thanks! :clap:

refi64:
node10 SDK extension has been merged, so this should be good to go!

gasinvein:
Great job @refi64, thanks!

PR #56, opened 2019-05-29
Node extras

refi64:
Missed this in the initial merge by a few hours :man_facepalming:, apparently a >100MB generated-sources.json is a normal thing in the npm world but isn't well-liked by GitHub, so I added a splitting mode, also documented some Chromedriver-related stuff.


barthalion:
Thanks!

PR #60, opened 2019-07-12
json2yaml: do not overwrite the yml file

muelli:
This prevents data loss.

I was accidentally executing the conversion script after I was nearly done. That caused my work to disappear. Sure, it's my fault, but I would have appreciated this safety net.

### DIFF
--- a/flatpak-json2yaml.py
+++ b/flatpak-json2yaml.py
@@ -78,6 +78,9 @@ def main():
         description='Flatpak JSON to YAML converter')
     parser.add_argument('json_file', type=str,
                         help='The flatpak JSON file to convert')
+    parser.add_argument('-f', '--force', dest='force',
+                        default=False, action='store_true',
+                        help='Overwrite existing file')
     parser.add_argument('-o', '--output', type=str, dest='out_file',
                         help='The yaml target path')
     args = parser.parse_args()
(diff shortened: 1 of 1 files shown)


TingPing:
I kind of like that behavior, maybe add a flag?

flatpak-json2yaml.py
```diff
@@ -89,8 +93,16 @@ def main():
...
         yaml_data = json_to_yaml(h.read())
 
-        with open(out_file, 'wb') as out:
-            out.write(yaml_data)
+        if args.force:
+            flags = O_CREAT|O_WRONLY|O_CLOEXEC
+        else:
+            flags = O_CREAT|O_EXCL|O_WRONLY|O_CLOEXEC
+
+        fd = os.open(out_file, flags)
```
bilelmoussaoui: I would have kept using context managers here instead of try/finally

flatpak-json2yaml.py
```diff
@@ -89,8 +93,16 @@ def main():
 
         yaml_data = json_to_yaml(h.read())
 
-        with open(out_file, 'wb') as out:
-            out.write(yaml_data)
+        if args.force:
+            flags = O_CREAT|O_WRONLY|O_CLOEXEC
+        else:
+            flags = O_CREAT|O_EXCL|O_WRONLY|O_CLOEXEC
```
refi64: This change isn't really needed, [Python has had a mode `x` for exclusive creation since 3.3](https://bugs.python.org/issue12760).

muelli: yeah, I also noticed. And I thought I had pushed that change, but I didn't.
Now i have, though.

flatpak-json2yaml.py
muelli: sure. it's not possible to do with `os.open` though.

PR #63, opened 2019-07-30
PIP: Prefer Generic Wheels over sdist

danyeaw:
## Overview
This PR changes the pip builder tool to use the normal preferred pip download (wheel if available and sdist if there is no binary), and only force pip to use the sdist if pip downloads a platform specific wheel. 

## Issue

In order to install an app using Poetry, I have to `pip install poetry`. Some of the Poetry dependencies like tomlkit require Poetry to build their sdist. So I get trapped in a catch-22, which can only be fixed by the FlatHub servers having Poetry (like setuptools), or we need to use the binary distributions that are available.

Use of a generic wheels should actually be preferred when they are available, since they offer:

- Faster installation for pure python packages
- Avoids arbitrary code execution for installation. (Avoids setup.py)
- Allows better caching for testing and continuous integration.
- Creates .pyc files as part of installation to ensure they match the python interpreter used.
- More consistent installs across platforms and machines.

My understanding why the pip flatpak-builder-tools don't use wheels is because generic wheels aren't available for all packages, and pip only allows options to set the platform to any (generic) if you force binary install using the `--binary-only :all:` option. This will then error out for any packages that don't have wheels.

## Solution

Updates the pip builder tool to conduct the following for each package:
1. Run a normal `pip download package`
2. Check for a platform specific wheel (the filename extension is not `all.whl` or `gz`)
3. If it is platform specific, delete the file and redownload the package, this time forcing `--no-binary`

I also updated the url for PyPI from pypi.python.org to pypi.org.

### DIFF
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -43,10 +43,16 @@ def get_pypi_url(name: str, filename: str) -> str:
 
 
 def get_package_name(filename: str) -> str:
-    segments = filename.split("-")
-    if len(segments) == 2:
-        return segments[0]
-    return "-".join(segments[:len(segments) - 1])
+    if filename.rsplit('.', 1)[1] == 'gz':
+        segments = filename.split("-")
+        if len(segments) == 2:
+            return segments[0]
+        return "-".join(segments[:len(segments) - 1])
+    elif filename.rsplit('-', 1)[1] == 'any.whl':
+        segments = filename.split("-")
+        if len(segments) == 5:
+            return segments[0]
+        return "-".join(segments[:len(segments) - 4])
 
 
 def get_file_hash(filename: str) -> str:
@@ -114,9 +120,21 @@ for package in packages:
             module['cleanup'] = ['/bin', '/share/man/man1']
 
         try:
-            subprocess.run(pip_download + [
-                '--no-binary', ':all:', package
-            ], check=True)
+            sdist_package = None
+            subprocess.run(pip_download + [package], check=True)
+            for filename in os.listdir(tempdir):
+                if filename.rsplit('.', 1)[1] != 'gz' and filename.rsplit('-', 1)[1] != 'any.whl':
+                    cwd = os.getcwd()
+                    os.chdir(tempdir)
+                    os.remove(filename)
+                    os.chdir(cwd)
+                    name = filename.split('-', 1)[0]
+                    sdist_package = name
+
+            if sdist_package:
+                subprocess.run(pip_download + [
+                    '--no-binary', ':all:', sdist_package
+                ], check=True)
             for filename in os.listdir(tempdir):
                 name = get_package_name(filename)
                 if name == 'setuptools':  # Already installed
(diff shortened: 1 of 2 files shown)


nedrichards:
Flathub is also pro source code by default since that is then automatically extracted and placed in a .Sources extension which helps with any licence obligations there may be. That may not be true for other sources of flatpaks.

danyeaw:
@nedrichards Wouldn't a Wheel satisfy this pro-source code view as well as a sdist? A wheel is a ZIP-format archive. It automatically includes the LICENSE files in the .dist-info directory, to support distribution of open source projects. The wheel also includes the full source code. So you get the full source, metadata about the project, and pre-compiled byte-code. I don't think it is a really a choice between having source code or not, because wheels contain everything that a sdist includes plus more, in order to improve speed and reproducibility.

nedrichards:
It would. But I seem to remember that flatpak doesn't see that as being source, although, solid point it will just get distributed in the main bundle so... yeah. Apologies for the irrelevant point.

pip/flatpak-pip-generator
```diff
@@ -114,9 +120,21 @@ for package in packages:
...
 
         try:
-            subprocess.run(pip_download + [
-                '--no-binary', ':all:', package
-            ], check=True)
+            sdist_package = None
+            subprocess.run(pip_download + [package], check=True)
+            for filename in os.listdir(tempdir):
+                if filename.rsplit('.', 1)[1] != 'gz' and filename.rsplit('-', 1)[1] != 'any.whl':
+                    cwd = os.getcwd()
```
TingPing: Whats with the directory dance?

pip/flatpak-pip-generator
```diff
@@ -114,9 +120,21 @@ for package in packages:
             module['cleanup'] = ['/bin', '/share/man/man1']
 
         try:
-            subprocess.run(pip_download + [
-                '--no-binary', ':all:', package
-            ], check=True)
+            sdist_package = None
+            subprocess.run(pip_download + [package], check=True)
+            for filename in os.listdir(tempdir):
+                if filename.rsplit('.', 1)[1] != 'gz' and filename.rsplit('-', 1)[1] != 'any.whl':
```
TingPing:
```python
if filename.endswith(('.gz', '-any.whl')):
```

?

pip/flatpak-pip-generator
```diff
@@ -114,9 +120,21 @@ for package in packages:
             module['cleanup'] = ['/bin', '/share/man/man1']
 
         try:
-            subprocess.run(pip_download + [
-                '--no-binary', ':all:', package
-            ], check=True)
+            sdist_package = None
+            subprocess.run(pip_download + [package], check=True)
```
TingPing: Comment here elaborating why its attempted twice.

pip/flatpak-pip-generator
danyeaw: Thanks for the feedback, I agree this wasn't very elegant. It is now using `os.path.join` instead of changing the directory and then changing it back.

pip/flatpak-pip-generator
danyeaw: Updated, I didn't know about the endswith function, but I should have assumed Python has something nice like this :smile:

pip/flatpak-pip-generator
danyeaw: Comment added :+1:

danyeaw:
Thanks @TingPing ♥️

PR #68, opened 2019-09-07
Add flatpak-cpan-generator

refi64:
It's kinda ugly but works pretty well, after a lot of research this is the best way I could come up with. (Some background is in README.md.)


TingPing:
(I don't think anybody here can review perl but it seems safe and I'll take your word on the rest)

PR #77, opened 2019-10-24
Poetry: Remove optional dependencies from generated sources

danyeaw:
Signed-off-by: Dan Yeaw <dan@yeaw.me>

# Overview
Poetry allows optional dependencies to be defined. For example, you can define an optional group of dependencies in your `pyproject.toml` that are needed to build project docs.

# Current Behavior
The current behavior of the poetry builder tool is that it pulls in all optional dependencies.

# Updated Behavior
Since optional dependencies shouldn't be needed to build an application, this PR excludes these dependencies.


danyeaw:
Thanks @barthalion!

PR #78, opened 2019-11-02
Support new poetry metadata format

danyeaw:
Poetry upgraded the metadata in the lock files in poetry 1.0.0b2. This change supports the new metadata format, while keeping compatibility with the old format for users using an older version.


barthalion:
Thanks!

PR #87, opened 2019-12-18
cargo: vendorise git dependencies

muelli:
With this change, the script behaves a bit more like "cargo vendor".
That was the intention of the script with "normal", i.e. crates.io
dependencies. cargo vendor places git dependencies in the same folder
and produces a .cargo-checksum.json file.  We try to do it similarly.

I needed to patch the cargo generator to build a flatpak for deltachat: https://github.com/flathub/chat.delta.desktop/pull/25
I'll be more than happy if these changes won't be required or if someone makes it work more properly.

### DIFF
--- a/cargo/flatpak-cargo-generator.py
+++ b/cargo/flatpak-cargo-generator.py
@@ -5,7 +5,7 @@ import base64
 import toml
 import json
 from urllib.parse import quote as urlquote
-from urllib.parse import urlparse, ParseResult
+from urllib.parse import urlparse, ParseResult, parse_qs
 import sys
 import argparse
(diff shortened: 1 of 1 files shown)


gdesmott:
Thanks @muelli this patch looks exactly to what I was looking for. :)

Do you have an example of project actually using it? Are you patching the `Cargo.toml` in the flatpak build process to use the vendored crates?

muelli:
https://github.com/flathub/chat.delta.desktop/ uses this.
I am not patching Cargo.toml.
IIRC I do `cargo vendor` before running the generator.
HTH!

PR #105, opened 2020-04-07
node: Option to download ffmpeg archive

gasinvein:
Some apps require electron-ffmpeg archive to be present alongside electron archive, so that cache layout would look like this:
```
flatpak-node/electron-cache
├── electron-v4.2.12-linux-x64.zip
├── ffmpeg-v4.2.12-linux-x64.zip
└── SHASUMS256.txt-4.2.12
```
This removes `--electron-non-patented-ffmpeg` option and adds `--electron-ffmpeg=` option with two possible values - `lib` and `archive`. The former will do the same that `--electron-non-patented-ffmpeg` currently does, while the later will instead add a file source with ffmpeg archive and place it alongside electron.

### DIFF
--- a/node/README.md
+++ b/node/README.md
@@ -189,11 +189,14 @@ things to note:
 
 Both of these cases are handled by the electron-webpack-quick-start example.
 
-### Non-patented ffmpeg
+### ffmpeg support
+
+If your app needs separate ffmpeg archive, add `--electron-ffmpeg=archive` option
+to flatpak-node-generator.
 
 By defualt, the ffmpeg that Electron ships with has proprietary codecs built in
 like AAC and H.264. If you don't need these, you can pass
-`--electron-non-patented-ffmpeg` to flatpak-node-generator. This will download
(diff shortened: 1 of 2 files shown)


node/README.md
```diff
@@ -189,11 +189,14 @@ things to note:
 
 Both of these cases are handled by the electron-webpack-quick-start example.
 
-### Non-patented ffmpeg
+### ffmpeg support
+
+If your app needs separate ffmpeg archive, add `--electron-ffmpeg=archive` option
```
refi64, line 194: Might be worth noting where the extra archive gets placed and why you'd need it (e.g. particular packages / use cases and the errors that might arise otherwise).

gasinvein: I'm not sure what and when exactly require it. Looks like it's an opt-in in electron-downloader, probably for apps that want to create patent-free distribution themselves. I've stumbled upon it while building Theia (VS Code fork).

PR #106, opened 2020-04-08
node: Add support for electron node-headers

### LINKED ISSUE #104: [node] node-headers from electron
Some node native deps require rebuilding them via node-gyp, which requires node-headers. Electron provides their own, e.g. [node-v4.2.12-headers.tar.gz](https://www.electronjs.org/headers/v4.2.12/node-v4.2.12-headers.tar.gz) for electron 4.2.12. Without it, native deps still could be rebuilt (with node headers installed in SDK extension), but may result in broken binaries, giving errors like this at run time:
```
 Error: The module '/app/main/resources/app/node_modules/native-keymap/build/Release/keymapping.node'
was compiled against a different Node.js version using
NODE_MODULE_VERSION 64. This version of Node.js requires
NODE_MODULE_VERSION 69. Please try re-compiling or re-installing
the module (for instance, using `npm rebuild` or `npm install`).
```
To my understanding, this happens with electron ≥ 4.0.3 with node 10, but other broken combinations possible.
This can be solved by adding node-headers source matching electron version manually and pointing node-gyp to it via `npm_config_nodedir` env var.
@refi64  Can it be automated in `flatpak-node-generator.py` with e.g. option like `--add-electron-node-headers`?

gasinvein:
Fixes #104
@refi64 Sorry for rushing this, if you're in progress of implementing this yourself - feel free to close this PR.

### DIFF
--- a/node/README.md
+++ b/node/README.md
@@ -189,6 +189,18 @@ things to note:
 
 Both of these cases are handled by the electron-webpack-quick-start example.
 
+### node-gyp and native dependencies
+
+Some node/electron versions are binary incompatible and require rebuilding of
+native node dependencies for electron. In offline mode, it may result in broken ABI.
+If you are seing errors like 
+`The module 'something.node' was compiled against a different Node.js version`,
+then pass `--electron-node-headers` option to flatpak-node-generator and set 
+`npm_config_nodedir` to `flatpak-node/node-gyp/electron-current`.
+
+Also you may want to run `electron-rebuild -f` to make sure that native dependencies
+were rebuilt.
+
 ### ffmpeg support
 
 If your app needs separate ffmpeg for matching electron version, add
--- a/node/README.md
+++ b/node/README.md
@@ -193,13 +193,10 @@ Both of these cases are handled by the electron-webpack-quick-start example.
 
 Some node/electron versions are binary incompatible and require rebuilding of
 native node dependencies for electron. In offline mode, it may result in broken ABI.
-If you are seing errors like 
+If you are seeing errors like 
 `The module 'something.node' was compiled against a different Node.js version`,
 then pass `--electron-node-headers` option to flatpak-node-generator and set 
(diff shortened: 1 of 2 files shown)


node/README.md
```diff
@@ -189,6 +189,18 @@ things to note:
...
+### node-gyp and native dependencies
+
+Some node/electron versions are binary incompatible and require rebuilding of
+native node dependencies for electron. In offline mode, it may result in broken ABI.
+If you are seing errors like 
+`The module 'something.node' was compiled against a different Node.js version`,
+then pass `--electron-node-headers` option to flatpak-node-generator and set 
+`npm_config_nodedir` to `flatpak-node/node-gyp/electron-current`.
+
+Also you may want to run `electron-rebuild -f` to make sure that native dependencies
```
refi64, line 201: Nit: where would I be running this? Does it go inside the manifest? After the install but before the build? Might be worth clarifying...

node/README.md
```diff
@@ -189,6 +189,18 @@ things to note:
 
 Both of these cases are handled by the electron-webpack-quick-start example.
 
+### node-gyp and native dependencies
+
+Some node/electron versions are binary incompatible and require rebuilding of
+native node dependencies for electron. In offline mode, it may result in broken ABI.
+If you are seing errors like
```
refi64, line 196: s/seing/seeing/

node/README.md
gasinvein: Hm. I'd rather remove this note - if native deps aren't rebuilt during `install` or `electron-builder` phase - it's an app flaw not specific to flatpak.

gasinvein: Also this is probably not needed for yarn at all - it'll always do rebuild unless `--ignore-scripts` was passed, and during that correct node-header will be picked up thanks to `npm_config_nodedir`, so further rebuilds are redundant.

PR #108, opened 2020-04-12
cargo: Properly support git repos with workspaces

gasinvein:
Kind of follow-up for #107 

In order to properly support git repos with workspace instead of single package, we need to parse `Cargo.toml`'s in there to determine what packages they contain.

Doing so, we don't need cached git repos (`cargo/git/db`) at all, so all related stuff can be removed, including siphash module.

This is tested on [Veloren](https://gitlab.com/veloren/veloren).

Possible improvements:
- [ ] Find a faster way to retrieve `Cargo.toml`'s from remote git repos than just cloning whole repo
- [ ] In generated-sources, avoid cloning whole repo for each package contained in it

### DIFF
--- a/cargo/flatpak-cargo-generator.py
+++ b/cargo/flatpak-cargo-generator.py
@@ -72,7 +72,6 @@ def get_git_sources(package):
             'replace-with': VENDORED_SOURCES,
         }
     }
-
     rev = parse_qs(urlparse(source).query).get('rev')
     branch = parse_qs(urlparse(source).query).get('branch', ['master'])
     if rev:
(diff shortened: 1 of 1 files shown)


gasinvein:
@barthalion @muelli Is this good enough to merge? Currently, it's rather slow on projects with many dependencies from git, but should be reliable enough. We could speed up git processing later on (see "Possible improvements").

muelli:
I've tried this branch for my usecase (deltachat) and it works: https://github.com/flathub/chat.delta.desktop/pull/34.
Thanks for cleaning it up!

gasinvein:
@muelli Huh? Isn't Delta Chat an electron app?

muelli:
yes. but it uses a Rust library.

Issue #103, opened 2020-03-29
[node] yarn --offline - node ./download-chromedriver.js - Error: getaddrinfo EAI_AGAIN github.com


Arxcis:
**About**
Hi! I am trying to package **BloomRPC** https://github.com/uw-labs/bloomrpc.
Here is my PR on `flathub` https://github.com/flathub/flathub/pull/1406
I have looked for existing issues like this, but I did not find any.

**Build options**
```yaml
  - name: bloomrpc
    sources:
      - type: git
        url: http://github.com/uw-labs/bloomrpc
        commit: '1.4.1'
      - generated-sources.json
      - type: script
        dest-filename: run.sh
        commands:
          - export TMPDIR="$XDG_RUNTIME_DIR/app/$FLATPAK_ID"
          - /app/BloomRPC/bloomrpc
    buildsystem: simple
    build-options:
      append-path: /usr/lib/sdk/node12/bin:/app/yarn/bin:$FLATPAK_BUILDER_BUILDDIR/flatpak-node/chromedriver
      env:
        ELECTRON_CACHE: /run/build/bloomrpc/flatpak-node/electron-cache
        CHROMEDRIVER_SKIP_DOWNLOAD: 'true'
        npm_config_nodedir: /usr/lib/sdk/node12
    build-commands:
      - HOME=$PWD yarn config --offline set yarn-offline-mirror $FLATPAK_BUILDER_BUILDDIR/flatpak-node/yarn-mirror
      - yarn --offline
      - sh flatpak-node/electron-builder-arch-args.sh
      - yarn run --offline package -- $ELECTRON_BUILDER_ARCH_ARGS
      # Max icon size for Flatpaks are 512x512. Having bigger icons results in build error
      - rm -rf build/files/share/icons/hicolor/1024x1024
      - cp -r build/files/* $FLATPAK_DEST
      - install -Dm 755 run.sh $FLATPAK_DEST/bin/run
```
**Output** 
```
➜  flathub git:(com.github.uwlabs.BloomRPC) ✗  flatpak-builder --repo=.localrepo --force-clean build com.github.uwlabs.BloomRPC.yaml
Emptying app dir 'build'
Downloading sources
warning: redirecting to https://github.com/uw-labs/bloomrpc/
Fetching git repo http://github.com/uw-labs/bloomrpc, ref refs/tags/1.4.1
warning: redirecting to https://github.com/uw-labs/bloomrpc/
[...]

PR #112, opened 2020-04-18
Many, many node improvements

refi64:
This has been cooking for quite some time now and has some big new features:

- Package metadata is now trimmed to remove all data except for the versions that are used. This results in *massive* size reductions in the generated sources files: for Scratch, it shrank from ~170MB to around ~8MB, or **a 95% size reduction**. Idea was originally from @TingPing.
- Many requests are now cached on disk. For the Scratch Flatpak, generating the sources with a filled cache reduced the runtime from around 2 minutes and 27 seconds to just 41 seconds, which is **a 72% reduction in time**. @muelli please try it and give feedback.

Bug fixes:

- node-chromedriver works now, @Arxcis @x80486 please try. Note that upstream `@electron/get` no longer supports ELECTRON_CACHE, so I was originally working on a much bigger overhaul. However, no electron-chromedriver that's been released is using the new `@electron/get`, and it's not clear that disregarding ELECTRON_CACHE isn't actually a bug, so I'm not bothering to handle that mess yet.
- Fixed Git dependencies for Yarn, @x80486 again please try with your repo and lmk if it works.
- https://github.com/flatpak/flatpak-builder-tools/commit/47ce4294195557c92b0e91cf4c2ca0ac80ba11b9 seems to have broken Git sources at least for NPM, so I sent a fix. @muelli please try this and make sure it still works with your original reproducer for the bug.

And some cleanup:

- YAPF is used to auto format the file.
- All the boolean CLI options were refactored into options classes, to make passing them around less repetitive.

### DIFF
--- a/node/.style.yapf
+++ b/node/.style.yapf
@@ -1,0 +1,4 @@
+[style]
+based_on_style = pep8
+column_limit = 90
+indent_dictionary_value = true
--- a/node/README.md
+++ b/node/README.md
@@ -29,7 +29,8 @@ get npm with electron-builder.
 ```
 usage: flatpak-node-generator.py [-h] [-o OUTPUT] [-r] [-R RECURSIVE_PATTERN]
                                  [--registry REGISTRY] [--no-devel]
-                                 [--no-aiohttp] [--retries RETRIES] [-P] [-s]
+                                 [--no-aiohttp] [--no-requests-cache]
+                                 [--retries RETRIES] [-P] [-s]
                                  [--node-chromedriver-from-electron NODE_CHROMEDRIVER_FROM_ELECTRON]
                                  [--electron-ffmpeg {archive,lib}]
                                  [--electron-node-headers]
@@ -54,6 +55,7 @@ optional arguments:
   --no-devel            Don't include devel dependencies (npm only)
   --no-aiohttp          Don't use aiohttp, and silence any warnings related to
                         it
+  --no-requests-cache   Disable the requests cache
   --retries RETRIES     Number of retries of failed requests
   -P, --no-autopatch    Don't automatically patch Git sources from
                         package*.json
@@ -73,6 +75,12 @@ containing all the sources set up like needed for the given package manager.
 
 If you're on npm and you don't want to include devel dependencies, pass --no-devel, and pass
 --production to `npm install` itself when you call.
+
+### Caching
+
+flatpak-node-generator will cache many API responses and archives from the server to speed up
+subsequent runs. You can disable this using `--no-requests-cache`, and it can be cleared via
+`rm -rf ${XDG_CACHE_HOME:-$HOME/.cache}/flatpak-node-generator`.
 
 ### Splitting mode
 
--- a/node/README.md
+++ b/node/README.md
@@ -30,8 +30,9 @@ get npm with electron-builder.
 usage: flatpak-node-generator.py [-h] [-o OUTPUT] [-r] [-R RECURSIVE_PATTERN]
                                  [--registry REGISTRY] [--no-devel]
                                  [--no-aiohttp] [--retries RETRIES] [-P] [-s]
-                                 [--electron-chromedriver ELECTRON_CHROMEDRIVER]
-                                 [--electron-non-patented-ffmpeg]
+                                 [--node-chromedriver-from-electron NODE_CHROMEDRIVER_FROM_ELECTRON]
+                                 [--electron-ffmpeg {archive,lib}]
+                                 [--electron-node-headers]
                                  {npm,yarn} lockfile
 
 Flatpak Node generator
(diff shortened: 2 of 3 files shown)


refi64:
The use of YAPF made the diff a tad larger than otherwise, so if you want to see just the other, more relevant changes: https://github.com/flatpak/flatpak-builder-tools/pull/112/files/fe8fe717f2682b06b437d7bcb6f27c7bf7147673..ba655338fc5f5089dafcdd9a3e3c70361f1d0d01

I'll merge this in once the bugs in particular are confirmed fixed by the various reports.

Arxcis:
@refi64  I can confirm that this PR fixes #103 for me

Arxcis:
I got a new issue though, because the building got further than before. Should I open a new issue for this?
```bash
Running: yarn --offline
yarn install v1.22.4
[1/4] Resolving packages...
[2/4] Fetching packages...
warning mini-css-extract-plugin@0.5.0: Invalid bin field for "mini-css-extract-plugin".
warning url-loader@1.1.2: Invalid bin field for "url-loader".
info fsevents@1.2.7: The platform "linux" is incompatible with this module.
info "fsevents@1.2.7" is an optional dependency and failed compatibility check. Excluding it from installation.
[3/4] Linking dependencies...
warning " > react-sortable-hoc@1.9.1" has unmet peer dependency "prop-types@^15.5.7".
warning " > extract-text-webpack-plugin@3.0.2" has incorrect peer dependency "webpack@^3.1.0".
warning "jest > jest-cli > @jest/core > jest-resolve-dependencies@24.3.0" has unmet peer dependency "jest-resolve@^24.1.0".
warning "jest > jest-cli > jest-config > jest-resolve@24.3.0" has unmet peer dependency "jest-haste-map@^24.0.0".
[4/4] Building fresh packages...
$ node internals/scripts/CheckNativeDep.js && electron-builder install-app-deps package.json
  • electron-builder version=20.38.5
  • loaded configuration file=package.json ("build" field)
  • installing production dependencies platform=linux arch=x64 appDir=/run/build/bloomrpc/app
Error: /usr/lib/sdk/node12/bin/node exited with code 1
Output:
yarn install v1.22.4
[1/4] Resolving packages...
[2/4] Fetching packages...
info If you think this is a bug, please open a bug report with the information provided in "/run/build/bloomrpc/app/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.

Error output:
warning You don't appear to have an internet connection. Try the --offline flag to use the cache for registry queries.

```

refi64:
I went looking through bloom-rpc's package.json files and electron-builder, it seems the latter will 
(bizarrely) always try to re-download everything if [the electron-builder app directory isn't the root directory](https://github.com/electron-userland/electron-builder/blob/v29.30.0/packages/electron-builder/src/cli/install-app-deps.ts#L61).

`app/package.json` seems to [already call `electron-rebuild` on post-install](https://github.com/uw-labs/bloomrpc/blob/master/app/package.json#L13), so you might be able to use sed or a patch file to remove the `electron-builder` command from `package.json` altogether. Failing that, you could patch package.json and add a [new entry inside the `build` section](https://github.com/uw-labs/bloomrpc/blob/master/package.json#L37) `"npmArgs": ["--offline"]` and see if that works.

I don't think there's any way to generically handle this case in flatpak-node-generator though; it's not uncommon to have to do app-specific tweaks in addition to the more general stuff this tool does.

x80486:
Nailed! :partying_face: :balloon: 

You fixed both issues I had (and probably some others). `dest-filename` has the correct filename; and I see that `chromedriver` is now in the list of dependencies in `generated-sources.json`.

> [This](https://github.com/x80486/flathub/tree/org.standardnotes.standardnotes/) is working with your repository, using the `node-fixes-2` branch. If you like to check, just run `update.sh` and the generated output will be correct.

Arxcis:
> I went looking through bloom-rpc's package.json files and electron-builder, it seems the latter will
> (bizarrely) always try to re-download everything if [the electron-builder app directory isn't the root directory](https://github.com/electron-userland/electron-builder/blob/v29.30.0/packages/electron-builder/src/cli/install-app-deps.ts#L61).

Woow. Nice catch:+1: I will do a patch. 

So, the issue is that there are two `package.json`-files:
- `package.json`
- `app/package.json`
```json
    "moduleDirectories": [
      "node_modules",
      "app/node_modules"
    ],
```
And both of them are running `electron-rebuild` post install. And I probably don't need to run it twice? And if I need to run it twice, I have to make sure to patch something, so that `yarn` i runs offline the second time as well.

refi64:
If no reply from @muelli soon about the package that reproduced the git prefix bug, I'll merge in a few days.

muelli:
it seems to work well enough for me for the packages I maintain :)

PR #113, opened 2020-04-24
dub: Make http requests asynchronously

gasinvein:
This makes the dub sources generator almost 10 times faster, e.g. reducing generation time for backgammon app from ~30 seconds to 3-5 seconds.

@Cogitri is it ok to add hard dependency on an external python module (`aiohttp`)?

### DIFF
--- a/dub/flatpak-dub-generator.py
+++ b/dub/flatpak-dub-generator.py
@@ -2,23 +2,25 @@
 
 __license__ = 'MIT'
 import json
-import urllib.request
 import urllib.parse
 import hashlib
 import logging
 import argparse
+import asyncio
+import aiohttp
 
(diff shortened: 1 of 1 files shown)


gasinvein:
I would appreciate if @refi64 could review this and point out possible mistakes, since I'm not very familiar with writing asynchronous code.

Cogitri:
> @Cogitri is it ok to add hard dependency on an external python module (aiohttp)?

I think so, most distros have it in their repos: https://repology.org/project/python:aiohttp/versions

gasinvein:
@refi64 Doesn't `asyncio.gather(*aws)` return results in the same order that arguments were?

refi64:
...yes it does, I probably should try to do PR reviews at 3am :sweat_smile:

PR #118, opened 2020-05-10
node: Handle dugite special source

gasinvein:
Yet another prebuilt binary downloader. This one is required by GitHub Desktop.


refi64:
Thanks! Merging manually so I could run yapf on the commit's changes.

PR #129, opened 2020-06-11
node: Implement rc-file source provider

gasinvein:
The problem:
`.yarnrc` / `.npmrc` can define custom node.js target, so that native node modules will build against it. But since we only read lock files, there is no way for us to determine which node-headers versions are required in this case.

The proposed solution:
Read rcfiles as well, and generate sources based on their content.

I'm not sure if I took the right approach to fit in nicely to the code structure, so if not - I hope @refi64 will point me to the right direction.
This adds several new abstractions:
- `NodeHeaders` - the node version item, similar to `Package`
- `NodeHeadersProvider` - implements generation of flatpak-builder source entries based on given `NodeHeaders`
- `YarnRCFileProvider` - `.yarnrc` file parser, extracts `NodeHeaders` items

### DIFF
--- a/node/flatpak-node-generator.py
+++ b/node/flatpak-node-generator.py
@@ -726,8 +726,25 @@ class LockfileProvider:
 
 
 class RCFileProvider:
+    RCFILE_NAME: str
+
+    def parse_rcfile(self, rcfile: Path) -> Dict[str, str]:
+        with open(rcfile, 'r') as r:
+            rcfile_text = r.read()
+        parser_re = re.compile(r'^(?!#|;)(\S+)(?:\s+|\s*=\s*)(?:"(.+)"|(\S+))$', re.MULTILINE)
+        result = {}
+        for key, quoted_val, val in parser_re.findall(rcfile_text):
+            result[key] = quoted_val or val
+        return result
+
     def get_node_headers(self, rcfile: Path) -> Optional[NodeHeaders]:
-        raise NotImplementedError()
+        rc_data = self.parse_rcfile(rcfile)
+        if 'target' not in rc_data:
+            return None
+        target = rc_data['target']
+        runtime = rc_data.get('runtime')
+        disturl = rc_data.get('disturl')
+        return NodeHeaders(target, runtime, disturl)
 
 
 class ModuleProvider(contextlib.AbstractContextManager):
@@ -1229,6 +1246,10 @@ class NpmLockfileProvider(LockfileProvider):
         yield from self.process_dependencies(lockfile, data.get('dependencies', {}))
 
 
+class NpmRCFileProvider(RCFileProvider):
+    RCFILE_NAME = '.npmrc'
+
+
 class NpmModuleProvider(ModuleProvider):
     class Options(NamedTuple):
         registry: str
@@ -1561,24 +1582,6 @@ class YarnLockfileProvider(LockfileProvider):
 class YarnRCFileProvider(RCFileProvider):
     RCFILE_NAME = '.yarnrc'
 
-    def parse_rcfile(self, rcfile: Path) -> Dict[str, str]:
-        parser_re = re.compile(r'^(\S+)\s+(?:"(.+)"|(\S+))', re.MULTILINE)
-        with open(rcfile, 'r') as r:
-            rcfile_text = r.read()
-        result = {}
-        for key, quoted_val, val in parser_re.findall(rcfile_text):
-            result[key] = quoted_val or val
-        return result
-
-    def get_node_headers(self, rcfile: Path) -> Optional[NodeHeaders]:
-        yarnrc = self.parse_rcfile(rcfile)
-        if 'target' not in yarnrc:
-            return None
-        target = yarnrc['target']
-        runtime = yarnrc.get('runtime')
-        disturl = yarnrc.get('disturl')
-        return NodeHeaders(target, runtime, disturl)
-
 
 class YarnModuleProvider(ModuleProvider):
     # From https://github.com/yarnpkg/yarn/blob/v1.22.4/src/fetchers/tarball-fetcher.js
@@ -1648,6 +1651,9 @@ class NpmProviderFactory(ProviderFactory):
 
     def create_lockfile_provider(self) -> NpmLockfileProvider:
         return NpmLockfileProvider(self.options.lockfile)
+
+    def create_rcfile_provider(self) -> NpmRCFileProvider:
+        return NpmRCFileProvider()
 
     def create_module_provider(self, gen: ManifestGenerator,
                                special: SpecialSourceProvider) -> NpmModuleProvider:
--- a/node/flatpak-node-generator.py
+++ b/node/flatpak-node-generator.py
@@ -524,6 +524,33 @@ class Package(NamedTuple):
     lockfile: Path
 
 
+class NodeHeaders:
+    def __init__(self, target: str, runtime: str = None, disturl: str = None):
+        self.target = target
+        if runtime is not None:
+            self.runtime = runtime
+        else:
+            self.runtime = 'node'
+        if disturl is not None:
+            self.disturl = disturl
+        elif self.runtime == 'node':
+            self.disturl = 'http://nodejs.org/dist'
+        elif self.runtime == 'electron':
+            self.disturl = 'https://www.electronjs.org/headers'
+        else:
+            raise ValueError(f'Can\'t guess `disturl` for {self.runtime} version {self.target}')
+
+    @property
+    def url(self) -> str:
+        #TODO it may be better to retrieve urls from disturl/index.json
+        return f'{self.disturl}/v{self.target}/node-v{self.target}-headers.tar.gz'
+
+    @property
+    def install_version(self) -> str:
+        #FIXME not sure if this static value will always work
+        return "9"
+
+
 class ManifestGenerator(contextlib.AbstractContextManager):
     MAX_GITHUB_SIZE = 49 * 1000 * 1000
     JSON_INDENT = 4
@@ -698,9 +725,34 @@ class LockfileProvider:
         raise NotImplementedError()
 
 
+class RCFileProvider:
+    def get_node_headers(self, rcfile: Path) -> Optional[NodeHeaders]:
+        raise NotImplementedError()
+
+
 class ModuleProvider(contextlib.AbstractContextManager):
     async def generate_package(self, package: Package) -> None:
         raise NotImplementedError()
+
+
+class NodeHeadersProvider:
+    def __init__(self, gen: ManifestGenerator):
+        self.gen = gen
+
+    @property
+    def gyp_dir(self) -> Path:
+        return self.gen.data_root / 'cache' / 'node-gyp'
+
+    async def generate_node_headers(self, node_headers: NodeHeaders):
+        url = node_headers.url
+        install_version = node_headers.install_version
+        dest = self.gyp_dir / node_headers.target
+        metadata = await RemoteUrlMetadata.get(url, cachable=True)
+        self.gen.add_archive_source(url,
+                                    metadata.integrity,
+                                    destination=dest)
+        self.gen.add_data_source(install_version,
+                                 destination=dest / 'installVersion')
 
 
 class ElectronBinaryManager:
@@ -1511,6 +1563,28 @@ class YarnLockfileProvider(LockfileProvider):
             yield self.parse_package_section(lockfile, section)
 
 
+class YarnRCFileProvider(RCFileProvider):
+    RCFILE_NAME = '.yarnrc'
+
+    def parse_rcfile(self, rcfile: Path) -> Dict[str, str]:
+        parser_re = re.compile(r'^(\S+)\s+(?:"(.+)"|(\S+))', re.MULTILINE)
+        with open(rcfile, 'r') as r:
+            rcfile_text = r.read()
+        result = {}
+        for key, quoted_val, val in parser_re.findall(rcfile_text):
+            result[key] = quoted_val or val
+        return result
+
+    def get_node_headers(self, rcfile: Path) -> Optional[NodeHeaders]:
+        yarnrc = self.parse_rcfile(rcfile)
+        if 'target' not in yarnrc:
+            return None
+        target = yarnrc['target']
+        runtime = yarnrc.get('runtime')
+        disturl = yarnrc.get('disturl')
+        return NodeHeaders(target, runtime, disturl)
+
+
 class YarnModuleProvider(ModuleProvider):
     # From https://github.com/yarnpkg/yarn/blob/v1.22.4/src/fetchers/tarball-fetcher.js
     _PACKAGE_TARBALL_URL_RE = re.compile(
@@ -1560,6 +1634,9 @@ class ProviderFactory:
     def create_lockfile_provider(self) -> LockfileProvider:
         raise NotImplementedError()
 
+    def create_rcfile_provider(self):
+        raise NotImplementedError()
+
     def create_module_provider(self, gen: ManifestGenerator,
                                special: SpecialSourceProvider) -> ModuleProvider:
         raise NotImplementedError()
@@ -1588,6 +1665,9 @@ class YarnProviderFactory(ProviderFactory):
 
     def create_lockfile_provider(self) -> YarnLockfileProvider:
         return YarnLockfileProvider()
+
+    def create_rcfile_provider(self) -> YarnRCFileProvider:
+        return YarnRCFileProvider()
 
     def create_module_provider(self, gen: ManifestGenerator,
                                special: SpecialSourceProvider) -> YarnModuleProvider:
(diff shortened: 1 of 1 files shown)


node/flatpak-node-generator.py
```diff
@@ -488,6 +488,39 @@ class Package(NamedTuple):
     lockfile: Path
 
 
+class NodeHeaders:
+    def __init__(self, target: str, runtime: str = None, disturl: str = None):
+        self.target = target
+        if runtime is not None:
+            self.runtime = runtime
+        else:
+            self.runtime = 'node'
```
refi64, line 497: maybe missing something obvious but, why not just `runtime: str = 'node'`?

node/flatpak-node-generator.py
```diff
@@ -619,11 +652,51 @@ def process_lockfile(self, lockfile: Path) -> Iterator[Package]:
...
+    def get_node_headers(self, rcfile: Path) -> NodeHeaders:
+        raise NotImplementedError()
+
+
 class ModuleProvider(contextlib.AbstractContextManager):
     async def generate_package(self, package: Package) -> None:
         raise NotImplementedError()
 
 
+class NodeHeadersProvider(contextlib.AbstractContextManager):
```
refi64, line 665: The AbstractContextManager in other classes was so that stuff could be done in `__exit__`, that's not used here so you can just drop the context manager / `__exit__` / the `with` statement later on

node/flatpak-node-generator.py
```diff
@@ -488,6 +488,39 @@ class Package(NamedTuple):
...
+
+    @property
+    def name(self):
+        return f'{self.runtime}-headers'
+
+    @property
+    def version(self):
+        return self.target
+
+    def get_url(self) -> str:
```
refi64, line 515: Nit: most `get_x(self)` functions would be better off as properties (also why the separate `def version()` property if it's identical to `target`? :thinking: )

node/flatpak-node-generator.py
```diff
@@ -619,11 +652,51 @@ def process_lockfile(self, lockfile: Path) -> Iterator[Package]:
...
+                                    metadata.integrity,
+                                    destination=dest)
+        self.gen.add_data_source(install_version,
+                                 destination=dest / 'installVersion')
+
+    async def generate_sdk_node_headers(self):
+        gyp_dir = self.get_gyp_dir()
+        install_version = "9"
+        script = f'''version="$(node --version | sed \'s/^v//\')"
+                     nodedir="$(dirname $(dirname $(which node)))"
```
refi64, line 691: Will this work if `enable.sh` is used? In that case, `node` wouldn't be in the path yet when the `shell` source commands are run... (tbh you could probably get away with just globbing for `/usr/lib/sdk/node*`, I'm not aware of any packages that actually need *multiple* versions of node at once)

node/flatpak-node-generator.py
```diff
@@ -619,11 +652,51 @@ def process_lockfile(self, lockfile: Path) -> Iterator[Package]:
...
+            dest = self.get_gyp_dir() / node_headers.target
+        metadata = await RemoteUrlMetadata.get(url, cachable=True)
+        self.gen.add_archive_source(url,
+                                    metadata.integrity,
+                                    destination=dest)
+        self.gen.add_data_source(install_version,
+                                 destination=dest / 'installVersion')
+
+    async def generate_sdk_node_headers(self):
+        gyp_dir = self.get_gyp_dir()
```
refi64, line 688: no need to be async here, also a bit confused on the naming. So we have generate_node_headers, which generates the targets to download node headers, but also generate_sdk_node_headers which doesn't actually download new headers? From what I can tell, this is symlinking the global SDK extension node headers for Gyp's use (based on the argument string), so maybe something about symlink_global_sdk_headers or similar, just for a bit more clarity!

node/flatpak-node-generator.py
```diff
@@ -1317,6 +1384,31 @@ def process_lockfile(self, lockfile: Path) -> Iterator[Package]:
             yield self.parse_package_section(lockfile, section)
 
 
+class YarnRCFileProvider(RCFileProvider):
+    rcfile_name = '.yarnrc'
```
refi64, line 1388: Nit, but could you use `ALL_CAPS` for consts? For consistency with the rest of the file (unless I missed one somewhere)

node/flatpak-node-generator.py
```diff
@@ -1317,6 +1384,31 @@ def process_lockfile(self, lockfile: Path) -> Iterator[Package]:
...
+
+    @classmethod
+    def parse_rcfile(cls, rcfile: Path) -> Dict[str, str]:
+        parser_re = re.compile(r'^(\S+)\s+(?:"(.+)"|(\S+))', re.MULTILINE)
+        with open(rcfile, 'r') as r:
+            rcfile_text = r.read()
+        result = {}
+        for key, quoted_val, val in parser_re.findall(rcfile_text):
+            result[key] = quoted_val + val
+        print(result)
```
refi64, line 1398: Leftover debug print?

node/flatpak-node-generator.py
```diff
@@ -1317,6 +1384,31 @@ def process_lockfile(self, lockfile: Path) -> Iterator[Package]:
             yield self.parse_package_section(lockfile, section)
 
 
+class YarnRCFileProvider(RCFileProvider):
+    rcfile_name = '.yarnrc'
+
+    @classmethod
+    def parse_rcfile(cls, rcfile: Path) -> Dict[str, str]:
```
refi64, line 1391: No need to be a classmethod from what I can tell

node/flatpak-node-generator.py
```diff
@@ -1317,6 +1384,31 @@ def process_lockfile(self, lockfile: Path) -> Iterator[Package]:
...
+        parser_re = re.compile(r'^(\S+)\s+(?:"(.+)"|(\S+))', re.MULTILINE)
+        with open(rcfile, 'r') as r:
+            rcfile_text = r.read()
+        result = {}
+        for key, quoted_val, val in parser_re.findall(rcfile_text):
+            result[key] = quoted_val + val
```
refi64, line 1397: would probably be clearer as `quoted_val or val` or similar, took me a moment to figure out the concat (also, does `.yarnrc` actually allow multiline strings?)

node/flatpak-node-generator.py
```diff
@@ -1355,10 +1447,16 @@ class ProviderFactory:
...
         raise NotImplementedError()
 
+    def create_rcfile_provider(self):
+        raise NotImplementedError()
+
     def create_module_provider(self, gen: ManifestGenerator,
                                special: SpecialSourceProvider) -> ModuleProvider:
         raise NotImplementedError()
 
+    def create_headers_provider(self, gen: ManifestGenerator) -> NodeHeadersProvider:
```
refi64, line 1457: ProviderFactory is mostly for stuff that's package-manager-dependent, I don't think the node headers really fit here? (though for the record rcfile provider probably does) If they do then this impl should probably be moved to NpmProviderFactory/YarnProviderFactory and then this is just a NotImplemented superclass

node/flatpak-node-generator.py
```diff
@@ -1355,10 +1447,16 @@ class ProviderFactory:
     def create_lockfile_provider(self) -> LockfileProvider:
         raise NotImplementedError()
 
+    def create_rcfile_provider(self):
```
refi64, line 1450: Won't this crash on npm because there's no override?

node/flatpak-node-generator.py
```diff
@@ -1580,6 +1693,12 @@ async def main() -> None:
         with provider_factory.create_module_provider(gen, special) as module_provider:
             with GeneratorProgress(packages, module_provider) as progress:
                 await progress.run()
+        with provider_factory.create_headers_provider(gen) as headers_provider:
+            if rcfile_node_headers:
+                with GeneratorProgress(rcfile_node_headers, headers_provider) as headers_progress:
```
refi64, line 1698: There are a few weird things here...

So GeneratorProgress in a lot of areas hardcodes text referring to packages, it also takes a ModuleProvider so I don't think this should actually type check. That being said, I don't think there would generally be enough header providers to justify using repurposing GeneratorProgress, maybe just a `print('something to do with headers')` or similar, since it will only be a few different cases I'd think. GeneratorProgress was largely added for npm because the generation can be so painfully slow there due to *so many metadata downloads*, lack of progress really makes it quite painful. But here, it would only ever be a few and in most cases one? so the time wouldn't be significant

node/flatpak-node-generator.py
```diff
@@ -761,19 +834,13 @@ def _handle_gulp_atom_electron(self, package: Package) -> None:
             self.gen.add_command(f'ln -sfTr "{self.electron_cache_dir}" "{cache_path}"')
 
     async def _handle_electron_headers(self, package: Package) -> None:
+        provider = NodeHeadersProvider(self.gen)
```
refi64, line 837: Might be a bit cleaner to pass this as a constructor argument, like we do with ManifestGenerator

node/flatpak-node-generator.py
```diff
@@ -1317,6 +1384,31 @@ def process_lockfile(self, lockfile: Path) -> Iterator[Package]:
...
+        for key, quoted_val, val in parser_re.findall(rcfile_text):
+            result[key] = quoted_val + val
+        print(result)
+        return result
+
+    def get_node_headers(self, rcfile: Path) -> NodeHeaders:
+        yarnrc = self.parse_rcfile(rcfile)
+        if 'target' not in yarnrc:
+            return None
+        nh_args = {'target': yarnrc['target']}
```
refi64, line 1405: I think this would be a bit nicer if it just assigned to some local variables that default to None (e.g. just `runtime = yarnrc.get('runtime')`) and passed them manually, it'll make it more resilient later on since mypy can actually check to make sure all the args match up.

node/flatpak-node-generator.py
gasinvein: That's for `GeneratorProgress()._format_package()` which expects package object with `name` and `version` attributes.
Hmm, maybe it should use `__str__()` representation of a package object instead?

node/flatpak-node-generator.py
gasinvein: The problem is that all installed SDK extensions are always added to sandbox. So, for local builds, multiple `node`'s are likely to be there.

node/flatpak-node-generator.py
gasinvein: Sorry, not sure if I understood. Should I remove `create_headers_provider()` altogether and use `NodeHeadersProvider` directly?

node/flatpak-node-generator.py
gasinvein: It will, `.npmrc` parsing is just not implemented yet.

node/flatpak-node-generator.py
gasinvein: Well, should we then just download collected node-headers in a loop?

node/flatpak-node-generator.py
gasinvein: Can you elaborate on this a bit, please?

node/flatpak-node-generator.py
gasinvein: Because we can't pass `runtime=None` to fallback to `node` otherwise.

node/flatpak-node-generator.py
```diff
@@ -619,11 +652,51 @@ def process_lockfile(self, lockfile: Path) -> Iterator[Package]:
...
+
+    async def generate_sdk_node_headers(self):
+        gyp_dir = self.get_gyp_dir()
+        install_version = "9"
+        script = f'''version="$(node --version | sed \'s/^v//\')"
+                     nodedir="$(dirname $(dirname $(which node)))"
+                     mkdir -p "{gyp_dir}/$version"
+                     ln -s "$nodedir/include" "{gyp_dir}/$version/include"
+                     echo "{install_version}" > "{gyp_dir}/$version/installVersion"'''
+        self.gen.add_command('\n'.join(l.strip() for l in script.splitlines()))
```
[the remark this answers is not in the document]
gasinvein: Would it? It'll be even longer and IMO just as hard to read.

node/flatpak-node-generator.py
refi64: That's definitely an oof. I don't really want to break enable.sh though, at least while it's still there, at minimum it should then be removed to avoid bizarre behavior. (Ideally we'd keep it since it's nice for debugging but...I legit have no idea how to handle this then. :neutral_face: ) Would it be *terrible* to go through all the directories since they'll have different versions anyway?

node/flatpak-node-generator.py
refi64: Is there anywhere where that's done though? Looking through this PR now, I see two places where NodeHeaders is used, neither of which `runtime` can be `None`. I'd rather generally avoid propagating `None`s for now if we can avoid it...

node/flatpak-node-generator.py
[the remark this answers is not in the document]
refi64: Well that's pretty much what the function is in the stdlib for... Actually in this case you wouldn't even need the `lstrip`, it would *just* be `self.gen.add_command(textwrap.dedent(script))`.

node/flatpak-node-generator.py
refi64: So, `ManifestGenerator` is provided as an argument to the constructor by the caller. There isn't really a major reason `NodeHeadersProvider` couldn't be the same way; it would be a bit cleaner and, most particularly, the main function is creating an instance *anyway*, might as well reuse it.

node/flatpak-node-generator.py
gasinvein: It comes from rcfile and could be `None`, here:
https://github.com/flatpak/flatpak-builder-tools/blob/9c43ee9196f94d35390d654ac78915cd8bf3bd6b/node/flatpak-node-generator.py#L1394-L1396

node/flatpak-node-generator.py
gasinvein: I guess going through all the directories has 50/50 chance to pick up wrong version. Maybe add a CLI option to set major node version explicitly, e.g. `--node-sdk-ext=node10`?

node/flatpak-node-generator.py
gasinvein: Hm, maybe remove `NodeHeadersProvider` altogether and move it's methods to `SpecialSourceProvider`?

node/flatpak-node-generator.py
gasinvein: Implemented a common parser that should work for both `.yarnrc` and `.npmrc`, so npm in now supported.

node/flatpak-node-generator.py
gasinvein: I've removed `NodeHeadersProvider` altogether and added its methods to `SpecialSourceProvider`. Seems to fit well there.

gasinvein:
From what I can tell, the only problem remains is that node headers generated from rcfiles can clash with the ones generated from lockfiles. This is _probably_ harmless, since flatpak-builder will just extract the same version tarball twice, but I'm not sure how bad it can be. Should we try to avoid this?

node/flatpak-node-generator.py
```diff
@@ -1171,6 +1229,10 @@ def process_lockfile(self, lockfile: Path) -> Iterator[Package]:
         yield from self.process_dependencies(lockfile, data.get('dependencies', {}))
 
 
+class NpmRCFileProvider(RCFileProvider):
+    RCFILE_NAME = '.npmrc'
```
refi64, line 1233: Having the two separate classes that only change a single constant on them feels a bit...off. It might be a bit cleaner if:

- There's one `RcFileProvider` that takes the file name as an instance variable.
  - Add two instance methods, `create_npm` and `create_yarn`, that do what it sounds like they do.
- Then `create_rcfile_providers` can just call those two methods, but only one actual class is needed.

node/flatpak-node-generator.py
```diff
@@ -690,6 +717,28 @@ def process_lockfile(self, lockfile: Path) -> Iterator[Package]:
         raise NotImplementedError()
 
 
+class RCFileProvider:
+    RCFILE_NAME: str
+
+    def parse_rcfile(self, rcfile: Path) -> Dict[str, str]:
```
refi64, line 723: Should be private I think.

node/flatpak-node-generator.py
```diff
@@ -1785,11 +1860,21 @@ async def main() -> None:
 
     print('Reading packages from lockfiles...')
     packages: Set[Package] = set()
+    rcfile_node_headers: Set[NodeHeaders] = set()
```
refi64, line 1863: I don't believe a set is actually going to work correctly. We can create a set of Packages because that's a NamedTuple, which defines the proper `__hash__` and `__eq__` methods for us. But, `NodeHeaders` doesn't have that, so this ends up functioning as a slightly more expensive list.

If de-duplicating values here is important, you could probably convert `NodeHeaders` to a `NamedTuple`, like:

```python
class NodeHeaders(NamedTuple):
    target: str
    disturl: str
    runtime: str = 'node'
```

then just add `NodeHeaders.with_guessed_disturl` or similar as a custom constructor that guesses the `disturl` value.

refi64 (changes requested):
Sorry if I missed this for a while! Not sure how long it hasn't been in WIP state anymore, it just got lost in my email I guess...

node/flatpak-node-generator.py
gasinvein: This is mostly for future. It's not unlikely that I've missed some npm/yarn-specific quirks that we'll need to handle.

node/flatpak-node-generator.py
gasinvein: Nah, it just parses rcflile to key-value mapping, without anything specific to the generator. It should have kind of "stable api".

node/flatpak-node-generator.py
gasinvein: Makes sense, will convert NodeHeaders to a NamedTuple subclass.

gasinvein: Done.

refi64 (approved):

refi64:
Holy crap, this has been open for a year now?? As usual, thanks for keeping up with all the different necessities coming up in the Node world!

PR #130, opened 2020-06-12
node: Create symlinks to zips for electron-get

gasinvein:
`@electron/get` is the new electron binary downloader, official replacement for `electron-download` package.
It doesn't support env vars that set path to electron downloads cache, and stores downloads in a subdirectory with name based on full download url.

### DIFF
--- a/node/flatpak-node-generator.py
+++ b/node/flatpak-node-generator.py
@@ -738,6 +738,13 @@ class SpecialSourceProvider:
                                     binary.integrity,
                                     electron_cache_dir / binary.filename,
                                     only_arches=[binary.arch.flatpak])
+            #Symlinks for @electron/get, which stores electron zips in a subdir
+            if self.xdg_layout:
+                sanitized_url = ''.join(c for c in binary.url if c not in '/:')
+                self.gen.add_shell_source(
+                    [f'ln -s "../{binary.filename}" "{binary.filename}"'],
+                    destination=electron_cache_dir / sanitized_url,
+                    only_arches=[binary.arch.flatpak])
 
         if add_integrities:
             integrity_file = manager.integrity_file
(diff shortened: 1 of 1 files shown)


node/flatpak-node-generator.py
```diff
@@ -606,6 +606,16 @@ def add_script_source(self, commands: List[str], destination: Path) -> None:
         source = {'type': 'script', 'commands': tuple(commands)}
         self._add_source_with_destination(source, destination, is_dir=False)
 
+    def add_shell_source(self,
+                         commands: List[str],
+                         destination: Optional[Path] = None,
+                         only_arches: Optional[List[str]] = None):
+        source = {'type': 'shell', 'commands': tuple(commands)}
```
refi64, line 613: We already have `add_command` to any commands needed once generation of everything else is complete (in `_finalize`), is there a particular reason for a separate add_shell_source? (esp. since the shell source doesn't actually take destination or arches, those will just be ignored, slightly surprised flatpak-builder didn't warn)

gasinvein: Yes, that's because of only_arches. Without it, we'll create a plenty of irrelevant broken symlinks. They're probably harmless, but it feels a bit messy.

refi64: Hmm, that's true. For some context, the reason scripts are all grouped together is just because the npm generator generates *so many*, it seriously kills performance. A cleaner solution might be to try to group scripts together but...that's a bit overkill for now given that there's only only one user... For now it should probably at least be `add_shell_source_slow` or similar w/ a warning comment to ideally not use it :sweat_smile:

gasinvein: Added method description suggesting to use `add_command()` instead.

refi64:
this'll work for now, thanks!

PR #132, opened 2020-06-29
Use requirement parser

### LINKED ISSUE #102: flatpak-pip-generator: Use existing parser
Use existing parsers like [setuptools](https://github.com/pypa/setuptools/blob/master/pkg_resources/_vendor/packaging/requirements.py) or [requirements-parser](https://pypi.org/project/requirements-parser/). The current parser seems to work only with exact package versions and not VCS specifiers, greater-than specifiers etc.

A6GibKm:
* Fixes #102
* Correctly names modules, e.g. occurrences of `"name": "python3-foo>"`
* If foo.txt is supplied output is named python3-foo.json
* For multiple modules output is named python3-modules.json
* Added instructions for multiple packages with versions in readme
* Tells the user the name of the output json file
* CVS generates a "git" type source, but won't build

### DIFF
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -47,15 +47,15 @@ def get_pypi_url(name: str, filename: str) -> str:
 
 def get_package_name(filename: str) -> str:
     if filename.endswith(('bz2', 'gz', 'xz', 'zip')):
-        segments = filename.split("-")
+        segments = filename.split('-')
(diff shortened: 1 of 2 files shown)


TingPing:
The style throughout the file is single quotes for strings.

TingPing:
And I do wish you split the changes into their own commits, but I won't block on that.

PR #135, opened 2020-07-03
pip: Fixes pip3 install command for vcs packages and extras

A6GibKm:
Tested with
```sh
# requirements.txt
git+git://github.com/psf/requests.git@c7e0fc087ceeadb8b4c84a0953a422c474093d6#egg=requests
certifi
```
produces a `json` file which is correctly builded with `flatpak-builder`. I am not convinced by the variable `name_for_pip` name.

### DIFF
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -118,6 +118,9 @@ for package in packages:
         ]
 
         name_for_pip = package.name
+        if len(package.extras) > 0:
+            extras = ', '.join(extra for extra in package.extras)
+            name_for_pip = name_for_pip + '[' + extras + ']'
         if package.vcs:
             name_for_pip = '.'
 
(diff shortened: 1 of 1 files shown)


pip/flatpak-pip-generator
```diff
@@ -116,14 +116,19 @@ for package in packages:
             '--dest',
             tempdir
         ]
+
+        name_for_pip = package.name
+        if package.vcs:
+         name_for_pip = '.'
```
TingPing, line 122: Bad whitespace.

A6GibKm: Good catch, fixed in a force push to have a single commit.

A6GibKm:
It was producing 
```json
"pip3 install --exists-action=i --no-index --find-links=\"file://${PWD}\" --prefix=${FLATPAK_DEST} foo"
```
instead of
```json
"pip3 install --exists-action=i --no-index --find-links=\"file://${PWD}\" --prefix=${FLATPAK_DEST} foo[bar1, bar2]"
```

PR #138, opened 2020-07-05
pip: Refactor how sources are generated 

A6GibKm:
DONE: ~~Find a way to list the sources files only once.~~
 with this PR the ALL sources will be listed once for each package. One possibility is to, instead of doing one module per package, do a single module with all packages. This would drastically decrease the length of the output file, even compared with `flatpak-pip-generator` before this PR, but we lose the ability to take off from the last built module in `flatpak-builder`.

### DIFF
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -79,7 +79,10 @@ def parse_continuation_lines(fin):
     for line in fin:
         line = line.rstrip('\n')
         while line.endswith('\\'):
-            line = line[:-1] + next(fin).rstrip('\n')
+            try:
+                line = line[:-1] + next(fin).rstrip('\n')
+            except StopIteration:
+                exit('Requirements have a wrong number of line continuation characters "\\"')
         yield line
 
 
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -139,7 +139,6 @@ output_filename = output_package + '.json'
 modules = []
 vcs_modules = []
 sources = {}
-vcs_sources = {}
 
 tempdir_prefix = 'pip-generator-{}'.format(output_filename.replace('.json', ''))
 with tempfile.TemporaryDirectory(prefix=tempdir_prefix) as tempdir:
@@ -224,14 +223,15 @@ with tempfile.TemporaryDirectory(prefix=tempdir_prefix) as tempdir:
                 ('url', url),
                 (s, revision),
             ])
-            vcs_sources[name] = {'source': source}
+            is_vcs = True
         else:
             url = get_pypi_url(name, filename)
             source = OrderedDict([
                 ('type', 'file'),
                 ('url', url),
                 ('sha256', sha256)])
-            sources[name] = {'source': source}
+            is_vcs = False
+        sources[name] = {'source': source, 'vcs': is_vcs}
 
 # Python3 packages that come as part of org.freedesktop.Sdk.
 system_packages = ['Mako', 'Markdown', 'meson', 'pip', 'setuptools', 'six']
@@ -281,15 +281,17 @@ for package in packages:
         except subprocess.CalledProcessError:
             print('Failed to download {}'.format(package.name))
 
-    package_sources = [sources[x]['source'] for x in dependencies if x in sources]
-    vcs_package_sources = [vcs_sources[x]['source'] for x in dependencies if x in vcs_sources]
+    is_vcs = True if package.vcs else False
+    package_sources = [sources[x]['source']
+                       for x in dependencies
+                       if x in sources
+                       # Add vcs source only if package is vcs
+                       and (not sources[x]['vcs'] or is_vcs)]
 
     if package.vcs:
         name_for_pip = '.'
-        vcs_s = vcs_package_sources
     else:
         name_for_pip = pkg
-        vcs_s = []
 
     module_name = 'python{}-{}'.format(python_version, package.name)
 
@@ -309,7 +311,7 @@ for package in packages:
         ('name', module_name),
         ('buildsystem', 'simple'),
         ('build-commands', [' '.join(pip_command)]),
-        ('sources', vcs_s + package_sources),
+        ('sources', package_sources),
     ])
     if opts.cleanup == 'all':
         module['cleanup'] = ['*']
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -83,11 +83,29 @@ def parse_continuation_lines(fin):
         yield line
 
 
+def parse_hashes(line):
+    name = line.split(" ")[0]
+    hash_requirements = []
+    while "--hash=" in line:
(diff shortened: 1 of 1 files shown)


A6GibKm:
Since `file` sources are downloaded only once when running `flatpak-builder`, the only drawback to this are:
* vcs modules will download still download more than once
* The output file will be harder to read due to the sources list being, probably gigantic.

A6GibKm:
This commit does `pip download package` on all packages (yes, packages can be downloaded up to 3 times, but most of the time they are on cache) to generate a dependency list and only list sources according to it. A small benchmark on 
```
pyaes>=0.1a1
ecdsa>=0.14
qrcode
protobuf>=3.12
dnspython<2.0
qdarkstyle<2.9
aiorpcx>=0.18,<0.19
aiohttp>=3.3.0,<4.0.0
aiohttp_socks>=0.3
certifi
bitstring
attrs>=19.2.0
```
Shows that even if `flatpak-pip-generator` does more calls of `pip download` the runtime %50 of what it was, and the output file is ~14% shorter, and only installs the correct versions.

pip/flatpak-pip-generator
```diff
@@ -74,12 +74,45 @@ def get_file_hash(filename: str) -> str:
...
+            if '--hash' in line:
+                use_hash = '--require-hashes'
+                package_name, hash_requirements = parse_hashes(line)
+                reqs[i] = package_name
+        reqs_as_str = "\n".join(reqs)
+        packages = list(requirements.parse(reqs_as_str))
+        requirements_file = opts.requirements_file
 elif opts.packages:
     packages = list(requirements.parse('\n'.join(opts.packages)))
+    requirements_file = '/tmp/flatpak-pip-generator-requirements.txt'
```
TingPing, line 116: Use the `tempfile` module.

pip/flatpak-pip-generator
```diff
@@ -74,12 +74,45 @@ def get_file_hash(filename: str) -> str:
...
+        hash_requirements.append([hash_algorithm.lower(), hash_value])
+    return name, hash_requirements
+
+
 packages = []
+use_hash = None
 if opts.requirements_file and os.path.exists(opts.requirements_file):
     with open(opts.requirements_file, 'r') as req_file:
-        packages = list(requirements.parse(req_file))
+        reqs = list(parse_continuation_lines(req_file))
```
TingPing, line 105: Why `list()`

pip/flatpak-pip-generator
```diff
@@ -74,12 +74,45 @@ def get_file_hash(filename: str) -> str:
...
+        for i, line in enumerate(reqs):
+            if '--hash' in line:
+                use_hash = '--require-hashes'
```
TingPing, line 108: Maybe turn this into a `bool` and then just insert the string below.

pip/flatpak-pip-generator
```diff
@@ -74,12 +74,45 @@ def get_file_hash(filename: str) -> str:
...
+                package_name, hash_requirements = parse_hashes(line)
+                reqs[i] = package_name
+        reqs_as_str = "\n".join(reqs)
+        packages = list(requirements.parse(reqs_as_str))
+        requirements_file = opts.requirements_file
 elif opts.packages:
     packages = list(requirements.parse('\n'.join(opts.packages)))
+    requirements_file = '/tmp/flatpak-pip-generator-requirements.txt'
+    with open(requirements_file, 'w') as req_file:
+        req_file.writelines('\n'.join(opts.packages))
```
TingPing, line 118: The point of `writelines` is it takes an iterable and you don't need to concat yourself?

pip/flatpak-pip-generator
```diff
@@ -74,12 +74,45 @@ def get_file_hash(filename: str) -> str:
...
+
+
+def parse_hashes(line):
+    name = line.split(" ")[0]
+    hash_requirements = []
+    while "--hash=" in line:
+        split = line.split("--hash=", 1)
+        hash_algorithm = split[1].split(":", 1)[0]
+        hash_value = split[1].split(":", 1)[1].split(" ", 1)[0]
+        line = line.replace('--hash=' + hash_algorithm + ":" + hash_value, '')
```
TingPing, line 96: Double quotes are used

pip/flatpak-pip-generator
A6GibKm: The temp file does not live enough, I will try it again tho.

A6GibKm: Fixed, there is a `delete=False` parameter.

pip/flatpak-pip-generator
A6GibKm: Done.

pip/flatpak-pip-generator
A6GibKm: Yes, but it does not add the newlines. But as you point, `write` is more apt.

pip/flatpak-pip-generator
A6GibKm: I am not happy with it either. What I do is to iter over `reqs` and change its contents (via `reqs[i]=foo`). Later `list()` is used again for the `packages` iterable so `len(packages) == 1` can be called without consuming the iterable, I haven't tough of a more graceful way of handling it.

pip/flatpak-pip-generator
```diff
@@ -74,15 +74,36 @@ def get_file_hash(filename: str) -> str:
...
 if opts.requirements_file and os.path.exists(opts.requirements_file):
     with open(opts.requirements_file, 'r') as req_file:
-        packages = list(requirements.parse(req_file))
+        reqs = parse_continuation_lines(req_file)
+        reqs_as_str = '\n'.join([r.split('--hash')[0] for r in reqs])
+        packages = list(requirements.parse(reqs_as_str))
+        requirements_file = opts.requirements_file
 elif opts.packages:
     packages = list(requirements.parse('\n'.join(opts.packages)))
+    with tempfile.NamedTemporaryFile('w', delete=False) as req_file:
```
TingPing, line 98: This does mean you leave temp files around forever which isn't ideal.

TingPing: Oh at the end you had a deletion check. You just need to store this object around and let it get deleted.

A6GibKm: Does this commit resolve the issue?

pip/flatpak-pip-generator
```diff
@@ -311,9 +311,8 @@ for package in packages:
     else:
         modules.append(module)
 
-if requirements_file == '/tmp/flatpak-pip-generator-requirements.txt':
-    if os.path.exists(requirements_file):
-        os.remove(requirements_file)
+if not opts.requirements_file and os.path.exists(requirements_file):
+    os.remove(requirements_file)
```
TingPing, line 315: This is a race condition. Just always call `os.remove()` and handle it failing as normal.

TingPing:
Thanks!

Issue #131, opened 2020-06-28
flatpak-pip-generator: kiwisolver-1.2.0 now throws ERROR: Command errored out with exit status 1


leinardi:
While generating the dependencies for `matplotlib==3.2.2`, `kiwisolver` throws this error:

```
Collecting matplotlib==3.2.2
  Using cached matplotlib-3.2.2-cp38-cp38-manylinux1_x86_64.whl (12.4 MB)
  Saved /tmp/pip-generator-python3-matplotlib-wis6876m/matplotlib-3.2.2-cp38-cp38-manylinux1_x86_64.whl
Collecting pyparsing!=2.0.4,!=2.1.2,!=2.1.6,>=2.0.1
  Using cached pyparsing-2.4.7-py2.py3-none-any.whl (67 kB)
  Saved /tmp/pip-generator-python3-matplotlib-wis6876m/pyparsing-2.4.7-py2.py3-none-any.whl
Collecting python-dateutil>=2.1
  Using cached python_dateutil-2.8.1-py2.py3-none-any.whl (227 kB)
  Saved /tmp/pip-generator-python3-matplotlib-wis6876m/python_dateutil-2.8.1-py2.py3-none-any.whl
Collecting kiwisolver>=1.0.1
  Using cached kiwisolver-1.2.0-cp38-cp38-manylinux1_x86_64.whl (92 kB)
  Saved /tmp/pip-generator-python3-matplotlib-wis6876m/kiwisolver-1.2.0-cp38-cp38-manylinux1_x86_64.whl
Collecting cycler>=0.10
  Using cached cycler-0.10.0-py2.py3-none-any.whl (6.5 kB)
  Saved /tmp/pip-generator-python3-matplotlib-wis6876m/cycler-0.10.0-py2.py3-none-any.whl
Collecting numpy>=1.11
  Using cached numpy-1.19.0-cp38-cp38-manylinux2010_x86_64.whl (14.6 MB)
  Saved /tmp/pip-generator-python3-matplotlib-wis6876m/numpy-1.19.0-cp38-cp38-manylinux2010_x86_64.whl
Collecting six>=1.5
  Using cached six-1.15.0-py2.py3-none-any.whl (10 kB)
  Saved /tmp/pip-generator-python3-matplotlib-wis6876m/six-1.15.0-py2.py3-none-any.whl
Successfully downloaded matplotlib pyparsing python-dateutil kiwisolver cycler numpy six
[...]

Issue #89, opened 2020-01-18
pip-generator fails for ipykernel


cvfosammmm:
After labouring for a while it fails with the following output. This is a blocker regards releasing my app as a flatpak, so it would be nice if it could be fixed.

```  Collecting pyzmq>=13 (from jupyter_client)
    Using cached https://files.pythonhosted.org/packages/3c/83/7ecbe9b762829f589fa9734026e0ccb63cb128fe5615ae1698f65df72bfe/pyzmq-18.1.1.tar.gz
  Collecting tornado>=4.1 (from jupyter_client)
    Using cached https://files.pythonhosted.org/packages/30/78/2d2823598496127b21423baffaa186b668f73cd91887fcef78b6eade136b/tornado-6.0.3.tar.gz
  Collecting parso>=0.5.2 (from jedi>=0.10->ipython>=5)
    Using cached https://files.pythonhosted.org/packages/e0/a2/3786c568fc8e9f64b9f7143e4c3904e61a8c5cb17260f22a6a3019d80c02/parso-0.5.2.tar.gz
  Collecting ptyprocess>=0.5 (from pexpect->ipython>=5)
    Using cached https://files.pythonhosted.org/packages/7d/2d/e4b8733cf79b7309d84c9081a4ab558c89d8c89da5961bf4ddb050ca1ce0/ptyprocess-0.6.0.tar.gz
    Installing build dependencies: started
    Installing build dependencies: finished with status 'error'
    Complete output from command /usr/bin/python3 -m pip install --ignore-installed --no-user --prefix /tmp/pip-build-env-pu0sb232 --no-warn-script-location --no-binary :all: --only-binary :none: -i https://pypi.org/simple -- flit:
    Collecting flit
      Using cached https://files.pythonhosted.org/packages/35/87/a92625dc5c81cf1608553b27b5c216bdc1e678ea6738a3a8a69ee2e5fb51/flit-2.2.0.tar.gz
      Installing build dependencies: started
      Installing build dependencies: finished with status 'error'
      Complete output from command /usr/bin/python3 -m pip install --ignore-installed --no-user --prefix /tmp/pip-build-env-pqaygza0 --no-warn-script-location --no-binary :all: --only-binary :none: -i https://pypi.org/simple -- "flit_core >=2.2.0,<3":
      Collecting flit_core<3,>=2.2.0
[...]

Issue #47, opened 2019-01-13
flatpak-pip-generator not working for matplotlib


leinardi:
If I generate the json module for `matplotlib` using `flatpak-pip-generator` and than I try to use it it fails with a numpy error:
```
Running: pip3 install --no-index --find-links="file://${PWD}" --prefix=${FLATPAK_DEST} matplotlib
Looking in links: file:///run/build/python3-matplotlib
Collecting matplotlib
Complete output from command python setup.py egg_info:
Download error on https://pypi.python.org/simple/numpy/: [Errno -3] Temporary failure in name resolution -- Some packages may not be found!
Couldn't find index page for 'numpy' (maybe misspelled?)
Download error on https://pypi.python.org/simple/: [Errno -3] Temporary failure in name resolution -- Some packages may not be found!
No local packages or working download links found for numpy>=1.10.0
```
If you manually add the command to install numpy if fails with `python-dateutil` not finding `setuptools`.

I managed to workaround this issue with this module configuration (not sure if the `sed` command is necessary):
```json
{
  "name": "python-setuptools_scm",
  "buildsystem": "simple",
  "build-commands": [
    "python3 setup.py install --prefix=/app --root=/"
  ],
  "sources": [
    {
      "type": "archive",
      "url": "https://files.pythonhosted.org/packages/09/b4/d148a70543b42ff3d81d57381f33104f32b91f970ad7873f463e75bf7453/setuptools_scm-3.1.0.tar.gz",
      "sha256": "1191f2a136b5e86f7ca8ab00a97ef7aef997131f1f6d4971be69a1ef387d8b40"
    },
    {
      "type": "shell",
      "commands": [
        "sed -i 's/if not platform_/if not dirs/' setup.py"
      ]
    }
  ]
}
```
```json
{
  "name": "python3-python-dateutil",
  "buildsystem": "simple",
  "build-commands": [
    "python3 setup.py install --prefix=/app --root=/"
  ],
  "sources": [
    {
      "type": "archive",
      "url": "https://files.pythonhosted.org/packages/0e/01/68747933e8d12263d41ce08119620d9a7e5eb72c876a3442257f74490da0/python-dateutil-2.7.5.tar.gz",
      "sha256": "88f9287c0174266bb0d8cedd395cfba9c58e87e5ad86b2ce58859bc11be3cf02"
    },
    {
      "type": "shell",
      "commands": [
        "sed -i 's/if not platform_/if not dirs/' setup.py"
      ]
[...]

Issue #26, opened 2018-06-24
pip-generator fails for pyqt5


ZanderBrown:
When running against the mu-editor package

```
Collecting pyqt5==5.10.1 (from mu-editor)
  Could not find a version that satisfies the requirement pyqt5==5.10.1 (from mu-editor) (from versions: )
No matching distribution found for pyqt5==5.10.1 (from mu-editor)
```
Or against pyqt5 directly
```
Collecting pyqt5
  Could not find a version that satisfies the requirement pyqt5 (from versions: )
No matching distribution found for pyqt5
Traceback (most recent call last):
  File "./flatpak-builder-tools/pip/flatpak-pip-generator", line 64, in <module>
    ] + opts.packages, check=True)
  File "/usr/lib64/python3.6/subprocess.py", line 418, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['pip3', 'download', '--dest', '/tmp/pip-generator-python3-pyqt5-kjvf0c4u', '--no-binary', ':all:', 'pyqt5']' returned non-zero exit status 1.
```

TingPing:
Looks like pyqt5 simply only ships binary releases on pypi, this script explicitly avoids binary releases since well.. that isn't portable or useful for many flatpaks.

The good news is dozens of packages build pyqt from source that you can use as an example: https://github.com/search?q=org%3Aflathub+pyqt&type=Code

ghisvail:
> this script explicitly avoids binary releases since well.. that isn't portable or useful for many flatpaks.

Agreed, though I'd still want to get an initial `generated-sources.json`, including the problematic `pyqt5` entry, to avoid writing them by hand. I can then remove the `pyqt5` entry and add it to the manifest separately using one of your examples.

Right now, it's all or nothing sadly.

ZanderBrown:
@ghisvail or just emit a warning and omit the entry? Largely anything but the current situation.

ghisvail:
@ZanderBrown, you can't really omit the entry, since you still need potential leaf dependencies for these packages which don't provide an appropriate source release on PyPI.

I tried hacking a solution by fiddling with the arguments baked to the `pip download` subprocess, using the `--only-binary` option followed by a selected set of dependencies (pyqt5, sip...). The script still fails further down the line for the applications I have tried (`mu-editor` and `spyder`).

ghisvail:
[...]

PR #139, opened 2020-07-06
pip: Eliminate calls of `pip download --no-binary`

A6GibKm:
This pr eliminates the need to run `pip` with the `--no-binary` flag, this causes a great speedup in the process and greately reduces the ammount of data to download. Files which are arch dependent are redownloaded using the info in pypi.org. Solves #47, #131, #89, and #26. But doesn't address the fact that some `tar.gz` packages can't be just installed with `pip install`, such as `pyqt5`.

Bonus: The output was prettified significantly and a warning message will be show if the python version used is not the same as of org.freedesktop.Platform, packages as `qdarkstyle` would be downloaded specifically for the host python version.

**Whats left to do in a future PR:**

* Add a `--only-host-arch` flag to allow the use of, for example `*x86_64.whl` packages, this would fix most of all remaining issues if the flatpak is built only for one arch only.

* Consider adding `setuptools` and `wheel` to all sources. It is almost random which packages will need this, and they are quite a few. Some packages invoke `pip install --ignore-installed`.

### DIFF
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -70,6 +70,7 @@ def get_package_name(filename: str) -> str:
             return segments[0]
         candidate = segments[:len(segments) - 4]
         # Some packages list the version number twice
+        # e.g. PyQt5-5.15.0-5.15.0-cp35.cp36.cp37.cp38-abi3-manylinux2014_x86_64.whl
         if candidate[-1] == segments[len(segments) - 4]:
             return '-'.join(candidate[:-1])
         return '-'.join(candidate)
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -6,7 +6,6 @@ import argparse
 import json
 import hashlib
 import os
-import requirements
 import shutil
 import subprocess
 import sys
@@ -149,6 +148,10 @@ if not opts.python2:
 
 packages = []
 if opts.requirements_file and os.path.exists(opts.requirements_file):
+    try:
+        import requirements
+    except ImportError:
+        exit('Requirements modules is not installed. Run "pip install requirements-parser"')
     with open(opts.requirements_file, 'r') as req_file:
         reqs = parse_continuation_lines(req_file)
         reqs_as_str = '\n'.join([r.split('--hash')[0] for r in reqs])
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -7,6 +7,7 @@ import json
 import hashlib
 import os
 import requirements
+import shutil
 import subprocess
 import sys
 import tempfile
@@ -46,6 +47,15 @@ def get_pypi_url(name: str, filename: str) -> str:
             raise Exception('Failed to extract url from {}'.format(url))
 
 
+def get_tar_package_url_pypi(name: str, version: str) -> str:
+    url = 'https://pypi.org/pypi/{}/{}/json'.format(name, version)
+    with urllib.request.urlopen(url) as response:
+        body = json.loads(response.read().decode('utf-8'))
+        for url in body['urls']:
+            if url['url'].endswith(('any.whl', 'gz', 'zip')):
+                return url['url']
+
+
 def get_package_name(filename: str) -> str:
     if filename.endswith(('bz2', 'gz', 'xz', 'zip')):
         segments = filename.split('-')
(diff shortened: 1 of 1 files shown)


A6GibKm:
Sample output: https://paste.centos.org/view/990020bd

TingPing:
I'll need to think about this more. Binaries are avoided for an obvious reason, they are built against a mystery ABI we have no clue if it works or not.

A6GibKm:
Previously the script did `pip download packages`, and for all packages that didn't end in `any.whl` or `tar.gz` it would run `pip download --no-binary` to get an arch independent version. Now instead of doing the second call of `pip download --no-binary` it will download the package directly from pypi in `tar.gz` or `any.whl` and delete the old file.

The only difference now is that instead of downloading in `tar.gz` format every single dependency of one of such packages, it would download the tar only for the packages that were arch dependent. Moreover pip, when called with `--no-binary` would try to compile each dependency and more often than not it would throw an error.

TingPing:
> Moreover pip, when called with --no-binary would try to compile each dependency

Yes that is the goal of flatpak.

> more often than not it would throw an error

Yes that sucks also.

A6GibKm:
> Yes that is the goal of flatpak.

At this point `pip` is running baremetal, this is inside `flatpak-pip-generator` and not `flatpak-builder`. Since no arch specific file will appear in the output manifest I am not sure if there will be any ABI incompability (I don't truly understand wheel or how python does things in this reggard).

pip/flatpak-pip-generator
```diff
@@ -56,16 +68,29 @@ def get_package_name(filename: str) -> str:
         segments = filename.split('-')
         if len(segments) == 5:
             return segments[0]
-        return '-'.join(segments[:len(segments) - 4])
+        candidate = segments[:len(segments) - 4]
+        # Some packages list the version number twice
```
TingPing, line 72: Just for ease of understanding maybe put an example.

pip/flatpak-pip-generator
```diff
@@ -86,6 +118,34 @@ def parse_continuation_lines(fin):
...
+    print(separator)
+
+
+if not opts.python2:
+    minor_version = str(sys.version_info[1])
+    command = [
+        'flatpak',
+        'run',
+        '--command=python3',
+        'org.freedesktop.Platform',
```
TingPing, line 134: We don't know this is the runtime they use.

pip/flatpak-pip-generator
```diff
@@ -86,6 +118,34 @@ def parse_continuation_lines(fin):
...
+        '--version'
+    ]
+    try:
+        result = subprocess.run(command, check=True, capture_output=True)
+        f_version = result.stdout.decode('utf-8')
+        f_minor_version = f_version.split('.')[1]
+        if minor_version != f_minor_version:
+            print('WARNING: the Python version on your host does not match the Python version in org.freedesktop.Platform')
```
TingPing, line 142: We can run all commands inside of `flatpak` actually, ignore the host entirely?

A6GibKm: Yes, it is possible. Will work on it for a future PR.

pip/flatpak-pip-generator
A6GibKm: I was thinking in doing something like: if there is a  `--runtime` flag use that, else try to use in this order `org.freedesktop.{Platform, Sdk}` and maybe add `org.{gnome,kde}.Platform`. Since this script is no longer compiling packages there is no need to use `Sdk` as first option.

I don't know if this feature should be left entirely optional and doing it baremetal if no runtime is found. It is a reasonable thing to assume the user will have some Platform runtime installed.

pip/flatpak-pip-generator
A6GibKm: I was thinking in doing something like: if there is a --runtime flag use that, else try to use in this order org.freedesktop.{Platform, Sdk} and maybe add org.{gnome,kde}.Platform. Since this script is no longer compiling packages there is no need to use Sdk as first option.

I don't know if this feature should be left entirely optional and doing it baremetal if no runtime is found. It is a reasonable thing to assume the user will have some Platform runtime installed.

pip/flatpak-pip-generator
A6GibKm: Fixed in a future pr.

pip/flatpak-pip-generator
A6GibKm: Added example for that line.

nanonyme:
Note that this is causing a horrifying amounts of issues with pip isolation currently as Python ecosystem is pushing strongly towards build isolation. I created https://github.com/flatpak/flatpak-builder-tools/pull/170 to try to restore sanity. It will more or less revert this.

PR #141, opened 2020-07-12
pip: Run inside of a sandbox

A6GibKm:
* For Python 3 packages it will run in the first runtime or sdk that it is available, running without a sandbox is not an option.

* Python 2 packages will use the host's `pip2` command unless a flatpak which bundles pip2 is specified. The is no natural choice flatpak the user should use as a default.

* The temporal requirements file is deleted as soon as it is not needed anymore.

### DIFF
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -158,6 +158,7 @@ else:
 if opts.runtime:
     flatpak_cmd = [
         'flatpak',
+        '--devel',
         '--share=network',
         '--filesystem=/tmp',
         '--command={}'.format(pip_executable),
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -35,7 +35,7 @@ parser.add_argument('--no-build-isolation',
 parser.add_argument('--output', '-o',
                     help='Specify output file name')
 parser.add_argument('--runtime',
-                    help='Specify a flatpak to run pip inside of a sandbox, ensures python version compatibility')
+                    help='Select a flatpak to run pip inside of a sandbox, ensures python version compatibility')
 opts = parser.parse_args()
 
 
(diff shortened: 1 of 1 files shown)


pip/flatpak-pip-generator
```diff
@@ -169,11 +150,65 @@ with open(requirements_file, 'r') as req_file:
     use_hash = '--hash=' in req_file.read()
 
 python_version = '2' if opts.python2 else '3'
+devnull = subprocess.DEVNULL
+
+# Prefer to use Sdk as they bundle `git`.
+possible_runtimes = [
```
TingPing, line 156: Not a big fan of this. We could import libflatpak which adds a dependency but makes it more robust...

Will have to think on that.

pip/flatpak-pip-generator
```diff
@@ -169,11 +150,65 @@ with open(requirements_file, 'r') as req_file:
     use_hash = '--hash=' in req_file.read()
 
 python_version = '2' if opts.python2 else '3'
+devnull = subprocess.DEVNULL
```
TingPing, line 153: Personal taste, not necessary.

pip/flatpak-pip-generator
```diff
@@ -169,11 +150,65 @@ with open(requirements_file, 'r') as req_file:
...
+
+    print('Using {} as a sandbox'.format(runtime))
+
 if opts.python2:
     pip_executable = 'pip2'
 else:
     pip_executable = 'pip3'
 
+if opts.python2 and not opts.runtime:
+    print('WARNING: --python2 requires a flatpak or runtime that bundles pip2 to be specified with --runtime for this script to be run inside a sandbox')
```
TingPing, line 196: Since no supported runtime contains `python2` maybe we should be more explicit that you should specify an application and not just runtimes.

pip/flatpak-pip-generator
```diff
@@ -169,11 +150,65 @@ with open(requirements_file, 'r') as req_file:
...
+    flatpak_cmd = [pip_executable]
+else:
+    flatpak_cmd = [
+        'flatpak',
```
TingPing, line 200: Pass `--devel` so the Sdk is used when possible.

A6GibKm: The issue solved by listing both Sdk and platform is that `flatpak run --devel org.freedesktop.Platform` does not have `git`.

pip/flatpak-pip-generator
A6GibKm: Changed to "flatpak app", might be better to say application.

pip/flatpak-pip-generator
A6GibKm: Me neither. Originally I wanted to use `org.freedesktop.Platform//19.08` as a default, but that requires maintenance both from the users an upstream. There is also the issue that if there are multiple branchs of the same runtime, `flatpak info --show-runtime $runtime` won't return anything.

I don't know if all distributions have access to libflatpak, at least silverblue provides it in the base image. I would consider this a regression if users need to install dependencies outside of python and there is no uniform way of providing them. I was thinking that in the future flatpak-builder-tools could be shipped as a flatpak.

pip/flatpak-pip-generator
TingPing: But I believe `flatpak run --devel org.example.App` will use the Sdk.

A6GibKm: On my machine I get
```
$ flatpak run --devel --command=git org.freedesktop.Platform
bwrap: execvp git: No such file or directory
$ flatpak run --command=git org.freedesktop.Sdk --version
git version 2.26.2
```

TingPing: Yes, app not runtime :P

pip/flatpak-pip-generator
```diff
@@ -192,28 +226,34 @@ modules = []
 vcs_modules = []
 sources = {}
 
-tempdir_prefix = 'pip-generator-{}'.format(output_filename.replace('.json', ''))
+tempdir_prefix = 'flatpak-pip-generator'
 with tempfile.TemporaryDirectory(prefix=tempdir_prefix) as tempdir:
-    pip_download = [
-        pip_executable,
+    pip_download = flatpak_cmd + [
```
gasinvein, line 231: This effectively makes it impossible to run this script in flatpak sandbox. At least add an option to run without flatpak, and/or check if we are already in sandbox.

A6GibKm: I didn't consider that case, will do.

A6GibKm:
I simplified the PR a bit. To avoid increasing the number of dependencies and possible errors, I made flatpak usage completely optional. The only notable change is that the temporal file now is deleted as soon as it is not needed anymore. We could use `with tempfile.TemporaryFile(...) as _`, which indeed would be better, but I don't feel like adding one level of indentation to most of the entire file.

A6GibKm:
cc @TingPing @gasinvein

Hofer-Julian:
Thanks @A6GibKm.
If with a later PR setuptools and wheel are added to all sources, it will finally allow us to update our dependencies with a single `
flatpak-pip-generator` run :)

A6GibKm:
I am not sure if just adding it to all sources is a good idea. I personally add it as a source manually from a file, see https://github.com/flathub/org.electrum.electrum. There are too many edge cases here, pyqt5, setuptools, wheel, in some cases even `pip` is required.

PR #144, opened 2020-07-15
node: Support for Yarn git packages

gasinvein:
Yarn mirrors git packages as uncompressed tarballs named after the  source repo and commit hash. I'm not quite sure if it's always the case, but it seems to be so for at least 1.22+

### DIFF
--- a/node/flatpak-node-generator.py
+++ b/node/flatpak-node-generator.py
@@ -1391,10 +1391,11 @@ class YarnModuleProvider(ModuleProvider):
             self.gen.add_url_source(source.resolved, integrity, self.mirror_dir / filename)
 
         elif isinstance(source, GitSource):
+            repo_name = urllib.parse.urlparse(source.url).path.split('/')[-1]
             tarball_url = source.tarball_url
             assert tarball_url is not None
             metadata = await RemoteUrlMetadata.get(tarball_url, cachable=True)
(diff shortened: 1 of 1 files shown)


refi64:
Out of curiosity, do you have an example lockfile for this? At the time I had first implemented this, yarn seemed to put the archive URL in the lock file, I guess that's changed?

gasinvein:
[Lots of them](https://github.com/search?q=filename%3Ayarn.lock+%22resolved%3A+git%2Bhttps%22).

gasinvein:
@muelli Do you maintain some npm-based apps? I'd appreciate if you could test this PR on them.

muelli:
I maintain [losslesscut](https://github.com/flathub/no.mifi.losslesscut/), a yarn-based app, and [delta-chat](https://github.com/flathub/chat.delta.desktop/), using npm.

With this branch, the generated sources for losslesscut do not change at all. For deltachat, ~8k lines have changed (many calls to `mkdir(indexv5/...)` which I guess could be written out in an ordered fashion). It still builds, so I guess it's working well enough.

gasinvein:
So it didn't break for at least git-less `package-lock.json` files? That's better than I expected, thanks.

gasinvein:
Now this is required for building VS Code (as of 1.48.0). @refi64, are you ok with this changes?

PR #146, opened 2020-07-23
cargo: maintain compatibility with Python 3.6

muelli:
The Python documentation states that "text" is a more readable alias for
"universinal_newlines". It was introduced with Python 3.7.
Ubuntu 18.04, however, run Python 3.6, is still supported, and doesn't
have an upgrade path yet.
This change brings compatibility with said Ubuntu version, but will
also benefit other Python 3.6 users.


muelli:
hm. Okay. I see your points. I have used decode() now.
If you think that's okay I'd squash the commits and merge them.

PR #147, opened 2020-07-26
cargo: Partial rewrite and more http usage

gasinvein:
This adds an option to download git-based crates as tarballs for known git hosts.

Adds `aiohttp` dependency.


gasinvein:
Since #149 it's pointless to download individual toml's via http (we need whole repos to use wildcards), so this functionality was removed.

PR #151, opened 2020-10-14
node: handle empty dependencies

muelli:
This change allows empty dependency declarations such as the one found
here: https://github.com/laurent22/joplin/blob/20bec7e26c175bc3fcd43be4c1a6c016c14a5fc8/Clipper/package.json

Passing the empty list should be semantically equivalent to no
dependencies.

It's probably not very common that a module doesn't have any dependencies... But I do wonder whether I am the first person to encounter empty dependencies.


refi64 (approved):
Thanks!

PR #155, opened 2020-11-03
[node] Generate our own tarball

Lctrs:
And stop relying on the ones generated by a git hosting provider.

Reason is that at least Github generates tarball with a root directory in the form of repo_name-commit_id/ which gets decompressed as is in node_modules by Yarn.
It is incompatible with webpack's resolver that expects a flat directory structure in node_modules.

### DIFF
--- a/node/flatpak-node-generator.py
+++ b/node/flatpak-node-generator.py
@@ -1385,10 +1385,10 @@ class YarnModuleProvider(ModuleProvider):
 
         elif isinstance(source, GitSource):
             repo_name = urllib.parse.urlparse(source.url).path.split('/')[-1]
-            repo_dir = self.gen.tmp_root / repo_name
-            target_tar = self.mirror_dir / f'{repo_name}-{source.commit}'
+            name = f'{repo_name}-{source.commit}'
+            repo_dir = self.gen.tmp_root / name
+            target_tar = self.mirror_dir / name
 
-            #XXX Yarn wants uncompressed tar, why?
             self.gen.add_git_source(source.url, source.commit, repo_dir)
             self.gen.add_command(f'cd {repo_dir}; git archive --format tar -o ../../../{target_tar} HEAD')
 
--- a/node/flatpak-node-generator.py
+++ b/node/flatpak-node-generator.py
@@ -1387,10 +1387,10 @@ class YarnModuleProvider(ModuleProvider):
             repo_name = urllib.parse.urlparse(source.url).path.split('/')[-1]
             name = f'{repo_name}-{source.commit}'
             repo_dir = self.gen.tmp_root / name
-            target_tar = self.mirror_dir / name
+            target_tar = os.path.relpath(self.mirror_dir / name, repo_dir)
 
             self.gen.add_git_source(source.url, source.commit, repo_dir)
-            self.gen.add_command(f'cd {repo_dir}; git archive --format tar -o ../../../{target_tar} HEAD')
+            self.gen.add_command(f'cd {repo_dir}; git archive --format tar -o {target_tar} HEAD')
 
         await self.special_source_provider.generate_special_sources(package)
 
--- a/node/flatpak-node-generator.py
+++ b/node/flatpak-node-generator.py
@@ -486,26 +486,6 @@ class GitSource(NamedTuple):
     commit: str
     from_: Optional[str]
 
-    @property
-    def tarball_url(self) -> Optional[str]:
-        url = urllib.parse.urlparse(self.url)
-        path = url.path.split('/')[1:]
-        #FIXME this is correct only for some git hosting providers
-        assert len(path) == 2, f'Can\'t get owner/repo from {self.url}'
-        owner = path[0]
-        if path[1].endswith('.git'):
-            repo = path[1].replace('.git', '')
-        else:
-            repo = path[1]
-        if url.hostname == 'github.com':
-            return f'https://codeload.{url.hostname}/{owner}/{repo}/tar.gz/{self.commit}'
-        elif url.hostname == 'gitlab.com':
-            return f'https://{url.hostname}/{owner}/{repo}/repository/archive.tar.gz?ref={self.commit}'
-        elif url.hostname == 'bitbucket.org':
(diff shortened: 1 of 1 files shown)


gasinvein (dismissed):
This code has nothing to do with webpack, it creates tarballs in a from that Yarn itself expects. If webpack expects something different - it should be handled separately.

Lctrs:
Yarn itself doesn't expect a tarball in this format. I did some tests by creating an offline mirror manually with yarn and then reinstall dependencies and check the differences with what I got with this script and it appeared that Yarn expects tarballs with no root directory.

Lctrs:
With my patch, node_modules was recreated as expected.

gasinvein (commented):
Looks like Yarn generates tarball without top-level directory indeed, but still handles fine tarballs with directory. Sorry, then.

Lctrs:
It doesn't handle tarball with root directory fine IMHO. It will just extracts it as is with no additional check because it expects only tarballs generated by itself (so with no root directory), which generates an invalid node_modules tree in our case.

gasinvein:
Well, it surely worked for me, not sure why. #144 was required to build VS Code 1.48.0.

node/flatpak-node-generator.py
```diff
@@ -1405,14 +1385,12 @@ async def generate_package(self, package: Package) -> None:
 
         elif isinstance(source, GitSource):
             repo_name = urllib.parse.urlparse(source.url).path.split('/')[-1]
-            tarball_url = source.tarball_url
-            assert tarball_url is not None
-            metadata = await RemoteUrlMetadata.get(tarball_url, cachable=True)
-            filename = f'{repo_name}-{source.commit}'
+            repo_dir = self.gen.tmp_root / repo_name
```
gasinvein, line 1388: It would be safer to add commit hash to the directory name, since there could be multiple versions of the same module.

node/flatpak-node-generator.py
```diff
@@ -1405,14 +1385,12 @@ async def generate_package(self, package: Package) -> None:
...
+            target_tar = self.mirror_dir / f'{repo_name}-{source.commit}'
 
             #XXX Yarn wants uncompressed tar, why?
-            self.gen.add_url_source(tarball_url, metadata.integrity, self.mirror_dir / f'{filename}.gz')
-            self.gen.add_command(f'gunzip "{self.mirror_dir}/{filename}.gz"')
+            self.gen.add_git_source(source.url, source.commit, repo_dir)
+            self.gen.add_command(f'cd {repo_dir}; git archive --format tar -o ../../../{target_tar} HEAD')
```
gasinvein, line 1393: ...or create tarball from specific commit here instead of HEAD.

Lctrs: The source is already at the commit, so HEAD points to the desired commit.

gasinvein: It won't necessarily be at correct commit if there are multiple instances of the same module with different commits.

Lctrs: Yeah but this will be fixed if we do what you said in your review above.

node/flatpak-node-generator.py
```diff
@@ -1405,14 +1385,12 @@ async def generate_package(self, package: Package) -> None:
...
         elif isinstance(source, GitSource):
             repo_name = urllib.parse.urlparse(source.url).path.split('/')[-1]
-            tarball_url = source.tarball_url
-            assert tarball_url is not None
-            metadata = await RemoteUrlMetadata.get(tarball_url, cachable=True)
-            filename = f'{repo_name}-{source.commit}'
+            repo_dir = self.gen.tmp_root / repo_name
+            target_tar = self.mirror_dir / f'{repo_name}-{source.commit}'
 
             #XXX Yarn wants uncompressed tar, why?
```
gasinvein, line 1391: I guess this comment is redundant now, as we don't have compressed tar anywhere anymore.

node/flatpak-node-generator.py
Lctrs: Adressed.

node/flatpak-node-generator.py
Lctrs: Removed.

node/flatpak-node-generator.py
```diff
@@ -1405,14 +1385,12 @@ async def generate_package(self, package: Package) -> None:
...
-
-            #XXX Yarn wants uncompressed tar, why?
-            self.gen.add_url_source(tarball_url, metadata.integrity, self.mirror_dir / f'{filename}.gz')
-            self.gen.add_command(f'gunzip "{self.mirror_dir}/{filename}.gz"')
+            name = f'{repo_name}-{source.commit}'
+            repo_dir = self.gen.tmp_root / name
+            target_tar = self.mirror_dir / name
+
+            self.gen.add_git_source(source.url, source.commit, repo_dir)
+            self.gen.add_command(f'cd {repo_dir}; git archive --format tar -o ../../../{target_tar} HEAD')
```
gasinvein, line 1393: It's more of a personal preference, but I'd rather calculate relative path at python level, e.g.
```suggestion
            repo_dir = self.gen.tmp_root / name
            target_tar = os.path.relpath(self.mirror_dir / name, repo_dir)

            self.gen.add_git_source(source.url, source.commit, repo_dir)
            self.gen.add_command(f'cd {repo_dir}; git archive --format tar -o {target_tar} HEAD')
```

Lctrs: Adressed.

gasinvein (approved):
LGTM, works for VS Code.

refi64:
So I'm gonna be honest, I'm mostly just confused as to why this would have worked until now without this, is it new Yarn behavior? Can you observe the same issues on older Yarn versions, i.e. that currently used by VS Code?

gasinvein:
@refi64 I have no idea. Yarn 1.22.4 creates tarballs without top-level directory.

Lctrs:
With root directory, Webpack can't resolve ES6 import.

refi64 (approved):
Well the code ends up shorter now, so I guess it's a net win? Thanks!

PR #158, opened 2020-11-18
cpan: Enable support for proxies when fetching archives

Saur2000:
This makes flatpak-cpan-generator.pl pick up, e.g., http_proxy and
https_proxy variables from the environment, allowing it to fetch
archives through a firewall.


refi64 (approved):
Thanks!

PR #159, opened 2020-11-18
cpan: Correct parsing of output from cpanminus

Saur2000:
cpanminus may output information about up/downgrading of modules it
installs, e.g.:

  Successfully installed PathTools-3.75 (upgraded from 3.74)

Correct the regular expression matching the installed packages so only
the name is matched.

### DIFF
--- a/cpan/flatpak-cpan-generator.pl
+++ b/cpan/flatpak-cpan-generator.pl
@@ -21,7 +21,7 @@ sub scan_deps {
 
   for (@deps)
   {
-      s/^Successfully installed (.*)/$1/;
+      s/^Successfully installed (\S+).*/$1/;
   }
 
    @deps


cpan/flatpak-cpan-generator.pl
```diff
@@ -21,7 +21,7 @@ sub scan_deps {
 
   for (@deps)
   {
-      s/^Successfully installed (.*)/$1/;
+      s/^Successfully installed (.*?)(?:\s+\(.*)?$/$1/;
```
refi64, line 24: If different things can be put after the package name, would it be easier to just do something like `(\S+).*`? That way this would work out on any other formats that cpanminus puts there.

Saur2000: Sure, it makes the regexp simpler as well. :)

Saur2000:
I'd forgotten about this. Anything else that needs to be done to have it merged?

muelli:
LGTM.
I have no interest in the Perl generator, but if nobody else feels entitled to merging this, I will.
Let's wait for a few days for anybody to oppose. Please nag me then.

Saur2000:
Well, it's been a week now. ;)

refi64:
Sorry for the delay (forgot about this), and @muelli thanks for merging!

PR #160, opened 2020-11-18
cpan: Make the output more deterministic

Saur2000:
Enable the canonical format for the JSON encoder. This avoids
unnecessary differences when committing the generated file in, e.g.,
git. Unfortunately the output is not completely deterministic as the
output from cpanm may vary.


muelli:
I don't know Perl, but LGTM

refi64 (approved):
Thanks again!

PR #164, opened 2020-12-04
pip: also skip dependency if it's a system package

Lctrs:


TingPing:
Should stop hardcoding system package names in the future.

PR #169, opened 2020-12-06
Reverse build isolation switch

### LINKED ISSUE #157: flatpak-pip-generator: unexpected behaviour for packages that rely on wheel
The flatpak-pip-generator produces incomplete json modules for packages that rely on wheel (not packaged as wheels, but list it as dependency).

**Steps to reproduce:**
Run
```
python flatpak-pip-generator.py debugpy
```
this outputs a file like this:
```json
{
    "name": "python3-debugpy",
    "buildsystem": "simple",
    "build-commands": [
        "pip3 install --exists-action=i --no-index --find-links=\"file://${PWD}\" --prefix=${FLATPAK_DEST} \"debugpy\""
    ],
    "sources": [
        {
            "type": "file",
            "url": "https://files.pythonhosted.org/packages/92/78/6f8857235b6b67194af3e66d0767d102e32a3e4055d0212f1783c0b4e5b0/debugpy-1.1.0.zip",
            "sha256": "955af25acd01ddd6a7adf6e495b118ed2adf072c61294a7e899c2c32f64e9a60"
        }
    ]
}
```
Adding this in a flatpak manifest fails with "ERROR: Could not find a version that satisfies the requirement wheel". Adding a module with "wheel" in front of it does not help. The only thing that works is adding "wheel" and "setuptools" sources to the module like this:
```json
{
      "name": "python3-debugpy",
      "buildsystem": "simple",
      "build-commands": [
        "pip3 install --exists-action=i --no-index --find-links=\"file://${PWD}\" --prefix=${FLATPAK_DEST}  \"debugpy\""
      ],
      "sources": [
        {
          "type": "file",
          "url": "https://files.pythonhosted.org/packages/92/78/6f8857235b6b67194af3e66d0767d102e32a3e4055d0212f1783c0b4e5b0/debugpy-1.1.0.zip",
          "sha256": "955af25acd01ddd6a7adf6e495b118ed2adf072c61294a7e899c2c32f64e9a60"
        },
        {
          "type": "file",
          "url": "https://files.pythonhosted.org/packages/a7/00/3df031b3ecd5444d572141321537080b40c1c25e1caa3d86cdd12e5e919c/wheel-0.35.1-py2.py3-none-any.whl",
          "sha256": "497add53525d16c173c2c1c733b8f655510e909ea78cc0e29d374243544b77a2"
        },
        {
          "type": "file",
          "url": "https://files.pythonhosted.org/packages/6d/38/c21ef5034684ffc0412deefbb07d66678332290c14bb5269c85145fbd55e/setuptools-50.3.2-py3-none-any.whl",
          "sha256": "2c242a0856fbad7efbe560df4a7add9324f340cf48df43651e9604924466794a"
        }
      ]
    }
```

nanonyme:
The reasonable default in pip that supports build isolation is to disable it. Fixes #157

### DIFF
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -30,8 +30,13 @@ parser.add_argument('--requirements-file', '-r',
 parser.add_argument('--build-only', action='store_const',
                     dest='cleanup', const='all',
                     help='Clean up all files after build')
-parser.add_argument('--no-build-isolation', action='store_true', default=False,
-                    help='https://pip.pypa.io/en/stable/reference/pip/#pep-517-and-518-support')
+parser.add_argument('--build-isolation', action='store_true',
+                    default=False,
+                    help=(
+                        'Do not disable build isolation. '
+                        'Mostly useful on pip that does\'t '
+                        'support the feature.'
+                    )
 parser.add_argument('--output', '-o',
                     help='Specify output file name')
 parser.add_argument('--runtime',
@@ -370,7 +375,7 @@ for package in packages:
         '--prefix=${FLATPAK_DEST}',
         '"{}"'.format(name_for_pip)
     ]
-    if opts.no_build_isolation:
+    if not opts.build_isolation:
         pip_command.append('--no-build-isolation')
 
     module = OrderedDict([


pip/flatpak-pip-generator
```diff
@@ -30,8 +30,6 @@ parser.add_argument('--requirements-file', '-r',
 parser.add_argument('--build-only', action='store_const',
                     dest='cleanup', const='all',
                     help='Clean up all files after build')
-parser.add_argument('--no-build-isolation', action='store_true', default=False
```
gasinvein, line 33: I'd rather invert the option instead of removing it completely. Just in case if someone would need it.

nanonyme: If it's imaginably desirable, then we should just drop the entire parameter and redesign the tool to support build isolation. It does not support it at the moment.

gasinvein: Oh. Now I wonder how it even worked if build isolation was enabled by default.

nanonyme:
Closing https://github.com/flatpak/flatpak-builder-tools/pull/170

pip/flatpak-pip-generator
nanonyme: It works mostly by luck.

nanonyme: Added parameter back but as reversed.

gasinvein:
I think there is no reason not to merge this. Pip with `--no-build-isolation` works for most cases when pip fails without it.

PR #170, opened 2020-12-06
pip: Generate single module with full sources

nanonyme:
This should work correctly with build isolation and fix #157
Essentially we generate a single module with a lot of sources, build wheels and finally install said wheels.

PR #167, opened 2020-12-06
pip: fix missing comma

HelpSeeker:

PR #172, opened 2020-12-09
pip: Fix missing comma in argparse function call

richardbarlow:


richardbarlow:
I've just seen that some other stuff is going on that means this will go away soon anyway (https://github.com/flatpak/flatpak-builder-tools/pull/170 and https://github.com/flatpak/flatpak-builder-tools/pull/167)

PR #175, opened 2020-12-13
fix: missing comma

tim77:


gasinvein:
Huh, it's the third time this fix is proposed (#167 and #172). It won't be needed after #170, but I guess if we don't merge it - it'll keep being submitted until #170 is merged.

Can you please also fix the script file permissions (set it to 755)?

gasinvein:
@nanonyme Please rebase #169 and #170

PR #184, opened 2021-02-10
node: Handle esbuild special source

gasinvein:
`esbuild` package has its native binaries in another npm packages, like `esbuild-linux-64`. The binary package is installed from `esbuild` post-install script and doesn't get into lockfile. Luckily, `esbuild` supports downloading package directly without using `npm` - we could use that.
Although I wonder if there is a way to somehow add to cache a package not listed in lockfiles - probably this could've been a better approach.

### DIFF
--- a/node/flatpak-node-generator.py
+++ b/node/flatpak-node-generator.py
@@ -960,6 +960,38 @@ class SpecialSourceProvider:
             self.gen.add_data_source("flatpak-node-cache",
                                      destination=destdir / 'INSTALLATION_COMPLETE')
 
+    async def _handle_esbuild(self, package: Package) -> None:
+        pkg_names = {
+            'x86_64': 'esbuild-linux-64',
+            'i386': 'esbuild-linux-32',
+            'arm': 'esbuild-linux-arm',
+            'aarch64': 'esbuild-linux-arm64'
+        }
+
+        for flatpak_arch, pkg_name in pkg_names.items():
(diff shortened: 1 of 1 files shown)


node/flatpak-node-generator.py
```diff
@@ -960,6 +960,38 @@ async def _handle_playwright(self, package: Package) -> None:
...
+            'i386': 'esbuild-linux-32',
+            'arm': 'esbuild-linux-arm',
+            'aarch64': 'esbuild-linux-arm64'
+        }
+
+        for flatpak_arch, pkg_name in pkg_names.items():
+            dl_url = f'https://registry.npmjs.org/{pkg_name}/-/{pkg_name}-{package.version}.tgz'
+            metadata = await RemoteUrlMetadata.get(dl_url, cachable=True)
+
+            archive_dst = self.gen.data_root / 'cache' / 'esbuild' / '.package' / f'{pkg_name}@{package.version}'
```
refi64, line 975: (very delayed, sorry!) nit, but couldn't this be `cache_dst / '.package' / ...` if the order is flipped?

gasinvein: Not sure if I get this. What's the difference? Should I define `cache_dst = self.gen.data_root / 'cache' / 'esbuild'`?

gasinvein: Done.

refi64 (approved):
Thanks!

PR #189, opened 2021-02-16
node: Add option to symlink SDK node headers

gasinvein:
This is needed in cases when native modules wirh node (not electron) ABI are required at build time (e.g. for node-sass) and setting `npm_config_nodedir` isn't enough.

Here we just symlink `/usr/lib/sdk/nodeN/include` to `flatpak-node/cache/node-gyp/VERSION/include`. Such script could be added manually, but I believe it would be more convinient if the generator handles it.

### DIFF
--- a/node/README.md
+++ b/node/README.md
@@ -69,6 +69,7 @@ optional arguments:
                         Download the ffmpeg binaries
   --electron-node-headers
                         Download the electron node headers
+  --xdg-layout          Use XDG layout for caches
 ```
 
 flatpak-node-generator.py takes the package manager (npm or yarn), and a path to a lockfile for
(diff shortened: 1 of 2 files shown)


refi64:
This LGTM, but could there be some documentation in the README as to when exactly someone might use this (i.e. the npm errors that would lead them to require it)?

gasinvein:
Hmm, maybe enable this by default instead? Node headers can't conflict with electron headers, and it they're not needed - they'll just be ignored.

refi64:
> Hmm, maybe enable this by default instead? Node headers can't conflict with electron headers, and it they're not needed - they'll just be ignored.

That sounds like a reasonable solution.

gasinvein:
Now sdk node headers are enabled by `--xdg-layout` option. Node-gyp headers are pretty useless otherwise anyway.

PR #191, opened 2021-03-09
Various minor Node fixes

refi64:
In short:

- Reformatting
- Type annotation fixes
- Fixes packages sometimes not working when there's a slash in the name
- Fixes ENOTCACHED errors when there's `/../` anywhere in the package's metadata

### DIFF
--- a/node/flatpak-node-generator.py
+++ b/node/flatpak-node-generator.py
@@ -1570,7 +1570,8 @@ class GeneratorProgress(contextlib.AbstractContextManager):
 
     async def run(self) -> None:
         self._update()
-        await asyncio.wait(map(self._generate, self.packages))
+        await asyncio.wait(
+            [asyncio.create_task(self._generate(pkg)) for pkg in self.packages])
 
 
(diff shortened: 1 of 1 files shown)


refi64:
Totally forgot this existed, merging now...

node/flatpak-node-generator.py
```diff
@@ -668,7 +688,7 @@ def process_lockfile(self, lockfile: Path) -> Iterator[Package]:
         raise NotImplementedError()
 
 
-class ModuleProvider(contextlib.AbstractContextManager):
+class ModuleProvider(contextlib.AbstractContextManager['ModuleProvider']):
```
gasinvein, line 691: Looks like this broke compatibility with Python ≤3.8 (including the one in fd.o sdk 20.08):
```
  File "../tools/node/flatpak-node-generator.py", line 691, in <module>
    class ModuleProvider(contextlib.AbstractContextManager['ModuleProvider']):
TypeError: 'ABCMeta' object is not subscriptable
```

PR #193, opened 2021-03-13
Add support for git+ssh scheme

proletarius101:
So naively adding this to the support list works.

### DIFF
--- a/node/flatpak-node-generator.py
+++ b/node/flatpak-node-generator.py
@@ -37,6 +37,7 @@ GIT_SCHEMES: Dict[str, Dict[str, str]] = {
     'git': {},
     'git+http': {'scheme': 'http'},
     'git+https': {'scheme': 'https'},
+    'git+ssh': {'scheme': 'ssh'},
 }
 GIT_URL_PATTERNS = [
     re.compile(r'^git:'),


muelli:
what problem does that fix for you?

proletarius101:
> what problem does that fix for you?

e.g. In https://github.com/bitwarden/desktop/blob/master/package.json#L309 they use `git+ssh`, which we would like to build from source: https://github.com/flathub/com.bitwarden.desktop/pull/52. This type of dependencies will abort the flatpak-builder-tools

muelli:
okay, cool! :) I haven't written the generator and I prefer somebody else taking the lead here for merging,  but if I merge this, I just want to make sure it fixes an actual problem rather than something that may be relevant in the future.

And so I wonder: does that really fix the problem? From what I can see the build fails in the download phase.

hfiguiere:
I don't think it should replace to `ssh:` as this require authtentication.

gasinvein:
`git+ssh` should work fine, flatpak-builder is able to handle it as-is. Can you provide at some evidence of what exactly does this fix?

muelli:
let me close for now. Please reopen if this becomes necessary.

PR #195, opened 2021-04-06
node: Support having multiple generated sources

joaquimrocha:
In some cases apps have different components/modules in the same
repository, and thus the flatpak for them would need several generated
node sources. However, if we try to use both generated sources in the
same flatpak module, they'd fail because we'd try to create directories
(index-v5 and hash-prefixes) that may exist.

Thus, in order to support using multiple generated sources, this patch
makes the generated sources be tolerant to the existence of the
mentioned folders.


muelli:
Looks innocent enough to me. Thanks!

PR #197, opened 2021-04-11
pip: add verbose flag to install command

Lctrs:
better to have some logs than a spin wheel that doesn't even work on flathub build system.

I also had the impression that it slightly improves build time, but I don't have any number to support this.

### DIFF
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -369,6 +369,7 @@ for package in packages:
     pip_command = [
         pip_executable,
         'install',
+        '--verbose',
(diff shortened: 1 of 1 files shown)


nanonyme:
I think this makes sense since we don't do the downloading part. pip install normally is overly verbose since it spams huge loads of download logs but here it should only give relevant output.

gasinvein:
Is it possible to also disable the progress spinner? It messes up log output when running non-interactively, e.g. in flathub builds.

Lctrs:
@gasinvein the verbose flag disables it.

PR #199, opened 2021-04-19
go-get: Add more known module URL replacements

joaquimrocha:
This patch adds some more replacements for known go modules' URLs.

### DIFF
--- a/go-get/flatpak-go-vendor-generator.py
+++ b/go-get/flatpak-go-vendor-generator.py
@@ -25,6 +25,8 @@ The real solution is https://github.com/golang/go/issues/35922
 import json
 import logging
 import sys
+import urllib.request
+from html.parser import HTMLParser
 
 import attr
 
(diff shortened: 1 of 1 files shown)


muelli:
LGTM.
I wonder: Can the list be found in the Go sources somewhere? maybe even re-used so that we don't have to maintain this list ourselves?

joaquimrocha:
I will try to find a way to get these automatically. Maybe checking the URL redirects is good enough.

joaquimrocha:
@muelli , I have updated the patch. It now follows a similar logic to the `go get` command.

joaquimrocha:
I forgot something important in the script. I'll fix it in a couple it hours. Don't merge.

joaquimrocha:
Should be fine now.

muelli:
works well for me (that is, syncthing).

It's a bit noisy for my taste, but that can be addressed later, if at all.

Thanks a million!

PR #206, opened 2021-06-01
dotnet: Handle multiple csproj files

marksvc:
- Allows generator to create a single sources list from multiple
.csproj files. A dependency may contain many .csproj files, but it
makes sense to group the sources into one .json file for the
dependency.
e.g. flatpak-dotnet-generator output.json library1.csproj
library2.csproj

### DIFF
--- a/dotnet/flatpak-dotnet-generator.py
+++ b/dotnet/flatpak-dotnet-generator.py
@@ -15,7 +15,7 @@ import tempfile
 def main():
     parser = argparse.ArgumentParser()
     parser.add_argument('output', help='The output JSON sources file')
-    parser.add_argument('project', help='The project file')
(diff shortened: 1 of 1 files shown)


marksvc:
I updated the PR to account for the change from dotnet5 to dotnet6.

dginovker:
Tested + works, saved me a bunch of time.

ProjectSynchro:
+1 to this as well, wish I had seen this before I went through the nightmare of merging JSON files :sweat_smile:

bbhtt:
Looks like this went through https://github.com/flatpak/flatpak-builder-tools/commit/c006b65247db1800fcc4676acfc4c06e402e27e7 Thanks for the PR.

PR #209, opened 2021-07-08
node: Update playwright special source handler

gasinvein:
- Download newer chromium revisions from MS CDN
- Skip browsers not installed by default
- Raise exception on unknown browser


refi64 (approved):
It's kind of hilarious to me how ridiculous the playwright code has become.

PR #214, opened 2021-07-17
cargo: Separate git repos from git packages

gasinvein:
Store loaded git repos in a special variable and pass it to package handler function, so that each repo is loaded only once.
Upon generating flatpak-builder sources, create git repo checkouts first, then, for each git package, copy subdir from checkout.

This should avoid unnecessary IO and make the generator behavior closer to what `cargo vendor` does.


tim77:
Tested this on https://github.com/flathub/net.veloren.veloren and LGTM so far.

gasinvein:
Veloren is probably the most complex rust-based app on Flathub. I guess we can assume that if it works for Veloren - it'll work for anything else.

Issue #202, opened 2021-05-04
flatpak-pip-generator fails on cyhunspell==2.0.2


Jenyay:
I run command 
``` 
python-pip-generator fails on
```
and I get an error
```
========================================================================
Downloading sources
========================================================================
Running: "pip3 download --exists-action=i --dest /tmp/pip-generator-python3-cyhunspellwqg93fyr -r /tmp/requirements.qgu20cuk"
/home/jenyay/.local/lib/python3.9/site-packages/secretstorage/dhcrypto.py:16: CryptographyDeprecationWarning: int_from_bytes is deprecated, use int.from_bytes instead
  from cryptography.utils import int_from_bytes
/home/jenyay/.local/lib/python3.9/site-packages/secretstorage/util.py:25: CryptographyDeprecationWarning: int_from_bytes is deprecated, use int.from_bytes instead
  from cryptography.utils import int_from_bytes
Collecting cyhunspell
  Using cached cyhunspell-2.0.2-cp39-cp39-manylinux2010_x86_64.whl (3.8 MB)
Collecting cacheman>=2.0.6
  Using cached CacheMan-2.1.0-py2.py3-none-any.whl (12 kB)
Collecting future>=0.16.0
  Using cached future-0.18.2.tar.gz (829 kB)
Collecting six>=1.10.0
  Using cached six-1.15.0-py2.py3-none-any.whl (10 kB)
Collecting psutil>=2.1.0
  Using cached psutil-5.8.0-cp39-cp39-manylinux2010_x86_64.whl (293 kB)
Saved /tmp/pip-generator-python3-cyhunspellwqg93fyr/cyhunspell-2.0.2-cp39-cp39-manylinux2010_x86_64.whl
Saved /tmp/pip-generator-python3-cyhunspellwqg93fyr/CacheMan-2.1.0-py2.py3-none-any.whl
Saved /tmp/pip-generator-python3-cyhunspellwqg93fyr/future-0.18.2.tar.gz
Saved /tmp/pip-generator-python3-cyhunspellwqg93fyr/psutil-5.8.0-cp39-cp39-manylinux2010_x86_64.whl
Saved /tmp/pip-generator-python3-cyhunspellwqg93fyr/six-1.15.0-py2.py3-none-any.whl
Successfully downloaded cyhunspell cacheman future psutil six
WARNING: You are using pip version 21.0.1; however, version 21.1.1 is available.
[...]

PR #217, opened 2021-08-11
Allow ignoring Pip packages

strugee:
This will allow generating the dep file and then manually fixing it up.

Ref #202

### DIFF
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -27,6 +27,8 @@ parser.add_argument('--cleanup', choices=['scripts', 'all'],
                     help='Select what to clean up after build')
 parser.add_argument('--requirements-file', '-r',
                     help='Specify requirements.txt file')
+parser.add_argument('--ignore-pkg', '-i', action='append',
+                    help='Ignore this package, e.g. if the package didn\'t upload source packages to PyPI')
 parser.add_argument('--build-only', action='store_const',
                     dest='cleanup', const='all',
                     help='Clean up all files after build')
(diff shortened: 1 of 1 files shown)


strugee:
Hrm, looks like for some reason packages I supposedly ignored on the commandline ended up in the final manifest, but using wheel sources? Not sure if that's a feature or a bug...

pip/flatpak-pip-generator
```diff
@@ -228,6 +230,8 @@ with tempfile.TemporaryDirectory(prefix=tempdir_prefix) as tempdir:
         if not filename.endswith(('bz2', 'gz', 'xz', 'zip')):
             version = get_file_version(filename)
             name = get_package_name(filename)
+            if  opts.ignore_pkg and name in opts.ignore_pkg:
+                continue
```
ENDrain, line 234: > Hrm, looks like for some reason packages I supposedly ignored on the commandline ended up in the final manifest, but using wheel sources? Not sure if that's a feature or a bug...

It's a feature. This condition is only applied during "Download arch independent packages"  stage. Prior to that wheels are downloaded at line 215/217. You will need to either interfere before that or at the manifest building stage

ENDrain (changes requested):

pip/flatpak-pip-generator
```diff
@@ -27,6 +27,8 @@ parser.add_argument('--cleanup', choices=['scripts', 'all'],
                     help='Select what to clean up after build')
 parser.add_argument('--requirements-file', '-r',
                     help='Specify requirements.txt file')
+parser.add_argument('--ignore-pkg', '-i', action='append',
+                    help='Ignore this package, e.g. if the package didn\'t upload source packages to PyPI')
```
ENDrain, line 31: Also, we shouldn't allow for `--ignore-pkg` and `packages` at the same time since that makes no sense. Parameter help should be updated accordingly

bbhtt:
I've posted an alternative take on this in https://github.com/flatpak/flatpak-builder-tools/pull/364

strugee:
@bbhtt I'll just close this in favor of your patches then? Realistically I'm not going to get around to fixing up this branch and it seems like you've got a working implementation anyway

strugee:
Thanks @bbhtt!

PR #225, opened 2021-09-04
node: Add NW.js support

gasinvein:
NW.js itself isn't stored in the lockfile (as far as I can tell), thus trigger NW.js source addition if the lockfile has `nw-builder` package - the most commonly used one to build NW.js apps.

I don't know a generic way to determine used NW.js version from well-known files, thus the NW.js version has to be passed on CLI via the new `--nw-version` option.

Unlike `electron/get`, `nw-builder` download cache can be stored anywhere (by default in the `cache/` subdir of the project), thus we download it to `flatpak-node/nwjs-cache/` dir, which should be symlinked manually to a proper path.

node-headers and 3rd-party ffmpeg are added for completeness, I'm not sure if these are useful in real projects.

### DIFF
--- a/node/flatpak-node-generator.py
+++ b/node/flatpak-node-generator.py
@@ -771,6 +771,7 @@ class SpecialSourceProvider:
         electron_node_headers: bool
         nwjs_version: str
         nwjs_node_headers: bool
+        nwjs_ffmpeg: bool
         xdg_layout: bool
 
     def __init__(self, gen: ManifestGenerator, options: Options):
@@ -780,6 +781,7 @@ class SpecialSourceProvider:
         self.electron_node_headers = options.electron_node_headers
         self.nwjs_version = options.nwjs_version
         self.nwjs_node_headers = options.nwjs_node_headers
+        self.nwjs_ffmpeg = options.nwjs_ffmpeg
         self.xdg_layout = options.xdg_layout
 
     @property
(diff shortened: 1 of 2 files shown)


node/flatpak-node-generator.py
```diff
@@ -900,6 +902,41 @@ async def _handle_node_chromedriver(self, package: Package) -> None:
...
+                                        metadata.integrity,
+                                        destination=dest,
+                                        only_arches=[flatpak_arch])
+
+    async def _handle_nw_builder(self, package: Package) -> None:
+        if self.nwjs_version:
+            version = self.nwjs_version
+        else:
+            versions_json = json.loads(await Requests.instance.read_all(
+                'https://nwjs.io/versions.json', cachable=False))
```
refi64, line 957: Assuming this would be used with nwbuild's `latest` version, is nwbuild actually able to download the latest without hitting the network every time? If it's not, it would probably be better to either drop this part, or add a note to the README saying to call the builder like `nwbuild -v $(cat blah-blah-blah/nwjs-version)` or similar.

node/flatpak-node-generator.py
```diff
@@ -1713,6 +1725,8 @@ async def main() -> None:
     parser.add_argument('--nwjs-node-headers',
                         action='store_true',
                         help='Download the NW.js node headers')
+    parser.add_argument('--nwjs-ffmpeg', action='store_true',
+                        help='Download prebuilt ffmpeg for current NW.js version')
```
refi64, line 1729: Might be worth also editing the help string `--electron-ffmpeg` above, since it's now a bit confusing.

refi64 (changes requested):
Looks almost perfect, just a few minor UX things I'd like to see done for clarity purposes.

node/flatpak-node-generator.py
gasinvein: Done.

node/flatpak-node-generator.py
gasinvein: Add NW.js note to readme. Hope it's clear enough.

refi64 (approved):

PR #230, opened 2021-09-24
node: Update dugite special source support

gasinvein:
- Dugite has at least one fork used in the wild, so we should handle it as well as the original package.
- Since GitHub repo may not have the tags matching given package version, pull package contents from a CDN instead of git repo

### DIFF
--- a/node/flatpak-node-generator.py
+++ b/node/flatpak-node-generator.py
@@ -64,6 +64,8 @@ GIT_URL_PATTERNS = [
 
 GIT_URL_HOSTS = ['github.com', 'gitlab.com', 'bitbucket.com', 'bitbucket.org']
 
+NPM_MIRROR = 'https://unpkg.com/'
+
 
 class Cache:
     instance: 'Cache'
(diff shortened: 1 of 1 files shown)


Lunarequest:
i can confirm this works as intended

node/flatpak-node-generator.py
```diff
@@ -1056,7 +1059,7 @@ async def generate_special_sources(self, package: Package) -> None:
             self._handle_electron_builder(package)
         elif package.name == 'gulp-atom-electron':
             self._handle_gulp_atom_electron(package)
-        elif package.name == 'dugite':
+        elif package.name in ['dugite', '@shiftkey/dugite']:
```
refi64, line 1062: minor nit: use a set comparison, it's slightly faster (but also a bit clearer, since sets are often used and thus associated with membership tests)

node/flatpak-node-generator.py
```diff
@@ -64,6 +64,8 @@
 
 GIT_URL_HOSTS = ['github.com', 'gitlab.com', 'bitbucket.com', 'bitbucket.org']
 
+NPM_MIRROR = 'https://cdn.jsdelivr.net/npm/'
```
refi64, line 67: nit: might be worth using unpkg if possible? Since it's used elsewhere, for consistency

refi64 (commented):
Just really minor nits, sole reason I'm not merging yet is because I believe it'll conflict with the NW.js PR due to the first commit

node/flatpak-node-generator.py
gasinvein: Makes sense, done.

node/flatpak-node-generator.py
gasinvein: Oh, I didn't notice a npm mirror is already used somewhere. Done.

refi64:
Thanks! Just did a rebase to fix the conflicts with the NW.js merge.

PR #231, opened 2021-09-27
remove depricated asyncio call

Lunarequest:
in python 3.10rc1 a deprecation warning is given for `asyncio.get_event_loop()` since there is no current event loop. instead a better way of starting a async block from synchronous code is to use `asyncio.run(function(args))`


refi64:
Well I was about to say that this won't work on 3.6, but apparently I accidentally broke 3.6 support earlier anyway :sweat_smile: it's losing security updates in 2 months, so that doesn't really matter.

PR #234, opened 2021-10-15
switch to semver over distutils

Lunarequest:
distutils is depreciated, PEP 632 recommends using `packaging` instead of `distultis.version`.

### DIFF
--- a/node/flatpak-node-generator.py
+++ b/node/flatpak-node-generator.py
@@ -9,7 +9,7 @@ from typing import *  # pyright: reportWildcardImportFromLibrary=false
 from typing import cast, IO
 
 from pathlib import Path
-from distutils.version import StrictVersion
+from semver import VersionInfo
 
 import argparse
 import asyncio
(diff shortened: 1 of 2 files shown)


node/flatpak-node-generator.py
```diff
@@ -9,7 +9,7 @@
 from typing import cast, IO
 
 from pathlib import Path
-from distutils.version import StrictVersion
+from packaging.version import parse
```
gasinvein, line 12: If we're going to hard depend on a 3rd-party module, I'd rather use `semver`.

nanonyme:
There should be package dependencies of this tool documented somewhere. Maybe requirements.txt?

gasinvein:
If add requirements.txt (and we now do have external dependencies), maybe flatpak-node-generator should hard depend on aiohttp and we should drop the urllib fallback?
But this is out of scope of this PR.

gasinvein:
@Lunarequest Nitpick: please squash your commits and add `node: ` prefix to the commit message (I can do it myself, but it'll remove your signature).

Lunarequest:
Will do.

node/requirements.txt
```diff
@@ -0,0 +1,10 @@
+aiohttp==3.7.4.post0
+async-timeout==3.0.1
+attrs==21.2.0
+chardet==4.0.0
+idna==3.3
+multidict==5.2.0
+requirements-parser==0.2.0
+semver==2.13.0
+typing-extensions==3.10.0.2
+yarl==1.7.0
```
gasinvein, line 10: The _node_ generator doesn't need most of these. If I'm not missing something, it only needs `aiohttp` and, now, `semver`. All the rest look like they belong to other generators.
And we certainly should not pin to exact versions (`==`).

refi64:
Oops I completely missed this PR before... Since then, the code has been massively reorganized and also [no longer depends on distutils anyway](https://github.com/flatpak/flatpak-builder-tools/blob/6f67a1fba4d9edd1e6d2e30fbb6e6f0fa2794186/node/flatpak_node_generator/package.py), so closing.

PR #235, opened 2021-10-23
Added support for whl extension in package url

TheJackiMonster:
I couldn't generate a dependency file with PyQt5 because the newest version only used .whl files. So the script always failed with an exception.

However adding the extension to the list which will be used for validation worked for me.


nanonyme:
This absolutely cannot be default behaviour. Some wheels are same (any wheel), they only have generated static metadata, otherwise they are equivalent to source packages. But other wheels have compiled code and they cannot be pulled in without user of tool consenting somehow.

nanonyme:
For compiled wheel even if they are acceptable, those have to be handled manually since you need different sources per architecture.

muelli:
so.. Do we close this as this cannot work reliably? Or do we give it the benefit of doubt, because it works sometimes?

TingPing:
As-is I don't think this should be merged. I think there is a middle ground of trying to use `any.whl`'s and/or making it explicitly opt-in (`--allow-whl-for=foo,bar`) or something.

TingPing:
Actually `any.whl` was supported in https://github.com/flatpak/flatpak-builder-tools/commit/fa73b5df4a46458085e93a929c4e507b3ae45a91

nanonyme:
Yeah, I the problem is things like Qt5 which indeed have wheels but they are per-arch and have ABI requirements. It's not a straightforward task to say what kind of manifest should be generated.

nanonyme:
Other unpleasant packages contains cryptography which assumes you have a Rust compiler if you want source install. But at least it can be done.

PR #240, opened 2021-11-01
node: Fix npm git patch from_ match

catsout:
From [npm-package-arg](https://github.com/npm/npm-package-arg#result-object), the `result.raw` may be `name + '@' + spec`.
In my test, I get:  
```text
package-lock.json:
npm6 : github:npm/hosted-git-info
npm7+: hosted-git-info@github:npm/hosted-git-info

package.json:
github:npm/hosted-git-info
```

From https://github.com/npm/arborist/blob/d9c603ede4fefb0bda1385e70b431b3adec810bf/lib/shrinkwrap.js#L961-L973
the `lock.from` normally be assigned `npa.resolve(node.name, edge.spec, edge.from.realpath).raw`

Just use regex to remove `name@` before real spec in `from_`,  as the `package.json` will not be like this.

### DIFF
--- a/node/flatpak-node-generator.py
+++ b/node/flatpak-node-generator.py
@@ -1377,7 +1377,8 @@ class NpmModuleProvider(ModuleProvider):
                     original_version = f'{source.original}'
                     new_version = f'{path}#{source.commit}'
                     assert source.from_ is not None
-                    data['package.json'][source.from_] = new_version
+                    from_ = re.sub(r'^[^@]+@git\+', 'git+', source.from_)
+                    data['package.json'][from_] = new_version
                     data['package-lock.json'][original_version] = new_version
 
                 for filename, script in scripts.items():


node/flatpak-node-generator.py
```diff
@@ -1377,7 +1377,8 @@ def _finalize(self) -> None:
                     original_version = f'{source.original}'
                     new_version = f'{path}#{source.commit}'
                     assert source.from_ is not None
-                    data['package.json'][source.from_] = new_version
+                    from_ = re.sub(r'^[^@]+@git\+', 'git+', source.from_)
```
refi64, line 1380: minor nit: it would be nice if this were `re.compile`'d in a constant, and I *think* you can avoid repeating `git+` in the replace arg by making the regex `^[^@]+(?=git\+)`

refi64 (commented):
Thanks for this! Largely looks good aside from a minor nit, but could you edit your commit message to have the text in the PR body? It makes it easier to track down changes in the future

PR #242, opened 2021-11-02
node: Use corgiDoc header to reduce size of generated-sources 

catsout:
From https://github.com/npm/pacote#packuments

```
curl https://registry.npmjs.org/node-emoji | jq
curl -H "Accept: application/vnd.npm.install-v1+json; q=1.0, application/json; q=0.8, */*"  https://registry.npmjs.org/node-emoji | jq
```
This is also the default behavior of npm

### DIFF
--- a/node/flatpak-node-generator.py
+++ b/node/flatpak-node-generator.py
@@ -247,18 +247,18 @@ class Requests:
 
     async def _read_parts(self,
                           url: str,
-                          size: int = DEFAULT_PART_SIZE) -> AsyncIterator[bytes]:
+                          size: int = DEFAULT_PART_SIZE, headers: dict[str,str] = None) -> AsyncIterator[bytes]:
         raise NotImplementedError
         yield b''  # Silence mypy.
 
-    async def _read_all(self, url: str) -> bytes:
+    async def _read_all(self, url: str, headers: dict[str,str] = None) -> bytes:
         raise NotImplementedError
 
     async def read_parts(self,
                          url: str,
                          *,
                          cachable: bool,
-                         size: int = DEFAULT_PART_SIZE) -> AsyncIterator[bytes]:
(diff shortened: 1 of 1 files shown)


node/flatpak-node-generator.py
```diff
@@ -247,18 +248,18 @@ def __get_cache_bucket(self, cachable: bool, url: str) -> Cache.BucketRef:
 
     async def _read_parts(self,
                           url: str,
-                          size: int = DEFAULT_PART_SIZE) -> AsyncIterator[bytes]:
+                          size: int = DEFAULT_PART_SIZE, headers: dict[str,str] = None) -> AsyncIterator[bytes]:
```
refi64, line 251: Two primary issues here:

- using the builtin types with `[]` syntax is invalid prior to Python 3.9, this should be `Dict[str, str]`
- default should be `{}`, `None` won't actually type check and also isn't entirely valid (I'm not sure if this worked due to sheer coincidence, but the defaults in urllib use `{}`, so relying on `None` working might not be guaranteed)

node/flatpak-node-generator.py
```diff
@@ -66,6 +66,7 @@
 
 NPM_MIRROR = 'https://unpkg.com/'
 
+NPM_CORGIDOC = 'application/vnd.npm.install-v1+json; q=1.0, application/json; q=0.8, */*'
```
refi64, line 69: Could you add a comment describing what this is / where it's from? I...honestly can';t find anything on a "corgidoc" header, unless my search-fu is just failing me :sweat_smile: (Unless it's just magic taken from somewhere in the npm source code, in which case...oof)

refi64 (changes requested):
Apparently I never noticed these 3 PRs were opened like...several months ago, sorry for the delay in reviewing!

PR #243, opened 2021-11-12
Update go-get readme to include working instructions

micahflee:
If you follow the instructions in the current readme in `go-get`, they don't work. Due to changes in Go, you now have to set the environment variable `GO111MODULE=off` for the build dir to end up in a format that `flatpak-go-get-generator.py` can work with.

I've added that, and also rewrote the instructions to make them more clear and easier to follow, with a full example manifest instead of just part of one and full example commands to run.

### DIFF
--- a/go-get/.gitignore
+++ b/go-get/.gitignore
@@ -1,0 +1,5 @@
+build
+.flatpak-builder
+example.json
+*-sources.json
+*-sources.yml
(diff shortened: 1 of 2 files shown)


muelli:
I think this has been addressed with https://github.com/flatpak/flatpak-builder-tools/pull/257

PR #246, opened 2021-11-18
Handle PEP-496 Environment Markers

nanonyme:
requirements-parser doesn't handle them internally so we need to
extract them from original line. Not handling these results in
unintended things being regarded as dependencies

### DIFF
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -11,6 +11,7 @@ import subprocess
 import sys
 import tempfile
 import urllib.request
+import platform
 
 from collections import OrderedDict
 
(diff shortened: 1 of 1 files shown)


nanonyme:
This isn't super-secure since it evaluates the trailer as code but it should showcase the problem and be useful as workaround if you review requirements.txt beforehand.

TingPing:
Maybe prompt before doing this?

nanonyme:
@TingPing it would not work with automatic updates.

nanonyme:
Technically right way probably is to use ast module to parse it and interpret the result manually making sure we bail out if something weird is noticed.

PR #248, opened 2021-12-07
Add x-checker-data to flatpak-pip-generator

JakobDev:
For use in [flatpak-external-data-checker](https://github.com/flathub/flatpak-external-data-checker).

Specific versions like foo==1.0 or foo<=1.0 does not work, because I don't know how to use them in x-data-checker.

### DIFF
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -2,6 +2,7 @@
 
 __license__ = 'MIT'
 
+from typing import Dict
 import argparse
 import json
 import hashlib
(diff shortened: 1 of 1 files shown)


pip/flatpak-pip-generator
```diff
@@ -129,6 +130,15 @@ def parse_continuation_lines(fin):
         yield line
 
 
+def get_pypi_checker_data(name: str, url: str) -> Dict[str, str]:
+    checker_data = {"type": "pypi", "name": name}
+    if url.endswith(".whl"):
+        checker_data["packagetype"] = "bdist_wheel"
+    elif url.endswith(".tar.gz"):
```
gasinvein, line 137: This does include all source file extension, you probably should check all of them, or just use `else:` here.

gasinvein (changes requested):
We really should pin versions by default, so auto-updating would be opt-in. Maintainer then can enable updating by changing exact version match to something else if they want.
In `x-checker-data` metadata, you can add a `versions` object containing something like `{"==": "1.2.0"}`.

JakobDev:
I personally think the x-checker-data should be included by default. The maintainer get only a notice that there are updates available. It's the choice of the maintainer if he/she want to use the updates. Updates can fix bugs and security problems, which is really important.

Another question: I can get the version range of a given package by using packages.but this does not include it's decencies. How can I get the version range of them? I have no idea.

JakobDev:
Support for x-checker-data was added in #256, so I'm closing this now

PR #250, opened 2021-12-17
node: Use inline type source to embed data

gasinvein:
Use the recently-added `inline` type source to embed arbitrary data instead of (ab)using `script` type source to generate a python script. This way it's a bit faster, makes generated sources file a bit smaller, and makes the flatpak-node-generator code much simpler.


refi64:
(Thank god, my previous solution for this was terrible.)

PR #260, opened 2022-03-08
flatpak-node-generator README: Mention XDG path of electron node headers

threema-danilo:
I had to dig in the source code to find out that the path is different when `--xdg-layout` is set.

### DIFF
--- a/node/README.md
+++ b/node/README.md
@@ -239,11 +239,28 @@ Both of these cases are handled by the electron-webpack-quick-start example.
 ### node-gyp and native dependencies
 
 Some node/electron versions are binary incompatible and require rebuilding of
-native node dependencies for electron. In offline mode, it may result in broken ABI.
-If you are seeing errors like
-`The module 'something.node' was compiled against a different Node.js version`,
-then pass `--electron-node-headers` option to flatpak-node-generator and set
-`npm_config_nodedir` to `flatpak-node/node-gyp/electron-current`.
+native node dependencies for electron. In offline mode, it may result in broken
+ABI. If you are seeing errors like `The module 'something.node' was compiled
+against a different Node.js version`, then pass `--electron-node-headers`
+option to flatpak-node-generator and set `npm_config_nodedir` to
+`flatpak-node/node-gyp/electron-current`.
+
+**Note**: Setting `npm_config_nodedir` should not be necessary when using XDG-compliant
+cache directories layout (the default, unless disabled via `--no-xdg-layout`).
+
+Some tools like *electron-rebuild* don't properly respect the
+XDG spec however. In this case, as a workaround, you might need to symlink the
+cache directory. For example:
+
+```yaml
+build-commands:
+  - |
+    ln -s $XDG_CACHE_HOME/node-gyp $HOME/.electron-gyp
+    npm run build
+```
+
+(Note that the build command must be ran as part of the same command as `ln`,
+i.e. it won't work if you run them as separate commands.)
 
 ### ffmpeg support
 


node/README.md
```diff
@@ -219,7 +219,8 @@ native node dependencies for electron. In offline mode, it may result in broken
 If you are seeing errors like
 `The module 'something.node' was compiled against a different Node.js version`,
 then pass `--electron-node-headers` option to flatpak-node-generator and set
-`npm_config_nodedir` to `flatpak-node/node-gyp/electron-current`.
+`npm_config_nodedir` to `flatpak-node/node-gyp/electron-current` (or
+`flatpak-node/cache/node-gyp/<version>` if you're using `--xdg-layout`).
```
gasinvein, line 223: Please make it clear that one doesn't need to set `npm_config_nodedir` if using `--xdg-layout`

threema-danilo: Is that the case? In my case I was already using `--xdg-layout`, but with `npm_config_nodedir: /usr/lib/sdk/node16` (which I took from some Electron-on-Flatpak tutorial, I think) the native libraries were (obviously, in hindsight) compiled against the wrong node headers. Setting `npm_config_nodedir` to the `flatpak-node/cache/node-gyp/<version>` matching my Electron version fixed the crashes.

If I remove the `npm_config_nodedir` variable completely, I get build errors when [electron-rebuild](https://github.com/electron/electron-rebuild) tries to fetch electron headers:

> Error: request to https://www.electronjs.org/headers/v16.0.7/node-v16.0.7-headers.tar.gz failed, reason: getaddrinfo EAI_AGAIN www.electronjs.org

I'm not sure if this is a problem of electron-rebuild, maybe it ignores the XDG path?

gasinvein: >Is that the case?

It is. The generator should've handled node-headers matching the node version used to build the app. If it didn't, then something must be ignoring the XDG path indeed.
In short, hardcoding `flatpak-node/cache/node-gyp/<version>` is not a proper solution and setting `npm_config_nodedir` is almost always a bad idea for several reasons.

What are you trying to build?

threema-danilo: > It is. The generator should've handled node-headers matching the node version used to build the app. If it didn't, then something must be ignoring the XDG path indeed.

The headers are there (and part of `generated-sources.json`), it works if I set `npm_config_nodedir`. So I assume it's `electron-rebuild` or `node-gyp` that ignores the XDG path. Do you have any link to specs/docs/code which specifies how/where the node headers should be looked up? Is this standardized, or do we just hope that the build tools won't change the lookup path in the future? If you have some links, maybe I can investigate and file an upstream issue.

> In short, hardcoding `flatpak-node/cache/node-gyp/<version>` is not a proper solution and setting `npm_config_nodedir` is almost always a bad idea for several reasons.

I assume one of the reasons is that it must be kept in sync with the `package.json` that pulls in Electron. Are there other reasons against it? (I'd definitely be happy if I can get rid of the hardcoded path.)

> What are you trying to build?

It's an internal Electron based application that is unfortunately not yet open source, so I cannot share the source code yet.

gasinvein: > > It is. The generator should've handled node-headers matching the node version used to build the app. If it didn't, then something must be ignoring the XDG path indeed.
> 
> The headers are there (and part of `generated-sources.json`), it works if I set `npm_config_nodedir`. So I assume it's `electron-rebuild` or `node-gyp` that ignores the XDG path. Do you have any link to specs/docs/code which specifies how/where the node headers should be looked up? Is this standardized, or do we just hope that the build tools won't change the lookup path in the future? If you have some links, maybe I can investigate and file an upstream issue.

Honestly, I'm not sure. There are so many native module building tools that I'm constantly confusing them. Most seem to use either the XDG-compliant `$XDG_CACHE_HOME/node-gyp/$version` path, or the legacy `$HOME/.node-gyp/$version`.
You can try using `strace` inside the build sandbox (`flatpak-builder --build-shell=your-app-module-name`) to figure out what path it tries to look for headers in.

> > In short, hardcoding `flatpak-node/cache/node-gyp/<version>` is not a proper solution and setting `npm_config_nodedir` is almost always a bad idea for several reasons.
> 
> I assume one of the reasons is that it must be kept in sync with the `package.json` that pulls in Electron. Are there other reasons against it? (I'd definitely be happy if I can get rid of the hardcoded path.)

Basically this. The other is reason is that the build workflow may require multiple node ABI versions (e.g. one dependency has to be built for node at build time, then another (or the same one) built for electron for deployment).

> > What are you trying to build?
> 
> It's an internal Electron based application that is unfortunately not yet open source, so I cannot share the source code yet.

Well, if you're not building the flatpak on Flathub infra (and don't plan to), you can just enable network access in the flatpak-builder sandbox and then don't need to use the generator at all.

threema-danilo: > Honestly, I'm not sure. There are so many native module building tools that I'm constantly confusing them.

I can relate to that 😂

I searched through the node-gyp issue tracker and found this: https://github.com/nodejs/node-gyp/pull/1570 So node-gyp itself _should_ look at the XDG based cache dir. And I'm setting `XDG_CACHE_HOME: /run/build/myproject/flatpak-node/cache`.

However – ughhh – it seems electron-rebuild doesn't use that directory: https://github.com/electron/electron-rebuild/blob/4d54c157816de564e3bc4f49f94e2139b231eeb5/src/constants.ts It uses `~/.electron-gyp` instead.

> Well, if you're not building the flatpak on Flathub infra (and don't plan to), you can just enable network access in the flatpak-builder sandbox and then don't need to use the generator at all.

I'm not totally sure yet, but we will probably publish a Flathub project. And it's certainly desireable to be able to know what's being downloaded during build time.

gasinvein: > However – ughhh – it seems electron-rebuild doesn't use that directory: https://github.com/electron/electron-rebuild/blob/4d54c157816de564e3bc4f49f94e2139b231eeb5/src/constants.ts It uses `~/.electron-gyp` instead.

You can try symlinking that path, then. In the manifest, do something like this:
```yaml
build-commands:
  - |
    ln -s $XDG_CACHE_HOME/node-gyp $HOME/.electron-gyp
    npm run build
```
Note that `npm run build` (or whatever your build stage command is) must be ran as part of the same command as `ln`, i.e. it won't work if you run them as consequent different commands.

threema-danilo: Great workaround, that did the trick!

Regarding this PR, what should I write? Should I update it with a note that the XDG variant shouldn't need `npm_config_nodedir` to be set, along with a hint about this issue?

gasinvein: > Regarding this PR, what should I write? Should I update it with a note that the XDG variant shouldn't need `npm_config_nodedir` to be set, along with a hint about this issue?

Adding a tip with this workaround would be fine. Ideally, electron-gyp should respect XDG spec, of course.
I personally believe we shouldn't advise to set `npm_config_nodedir` at all and make `--xdg-layout` the default, but it's kinda out of scope of this PR.

refi64: I'd be okay with making `--xdg-layout` the default FTR, maybe just throw out a deprecation warning when it's not given for a while & add a corresponding `--no-xdg-layout` if compatibility is a concern here? (This feels deja-vu, I can't remember if I already said this before, or I meant to say it and forgot :sweat_smile: )

threema-danilo: @gasinvein I updated the PR. What do you think about this version?

gasinvein: @threema-danilo  LGTM now, thanks.

MrMYHuang:
Thank you. It works.
However, by my test, the `ln` command line should be run before `npm install --offline`.
It could be related to https://www.electronjs.org/docs/latest/tutorial/using-native-node-modules
```
# Install all dependencies, and store cache to ~/.electron-gyp.
HOME=~/.electron-gyp npm install
```

This is my flatpak manifest for reference:
https://github.com/flathub/io.github.mrmyhuang.cbetar2/blob/7c1e6c0c3f0084574398a9d17109630f28ba007e/io.github.mrmyhuang.cbetar2.yml

MrMYHuang:
> However, by my test, the `ln` command line should be run before `npm install --offline`.

Sorry, this was wrong. I thought it was so, because my `npm install --offline` triggers the package.json postinstall script to rebuild native module by electron-builder.

Actually, by me tests, the `ln` command line shall be run before electron-builder rebuilds native modules.

node/README.md
```diff
@@ -215,11 +215,26 @@ Both of these cases are handled by the electron-webpack-quick-start example.
...
-`The module 'something.node' was compiled against a different Node.js version`,
-then pass `--electron-node-headers` option to flatpak-node-generator and set
-`npm_config_nodedir` to `flatpak-node/node-gyp/electron-current`.
+native node dependencies for electron. In offline mode, it may result in broken
+ABI. If you are seeing errors like `The module 'something.node' was compiled
+against a different Node.js version`, then pass `--electron-node-headers`
+option to flatpak-node-generator and set `npm_config_nodedir` to
+`flatpak-node/node-gyp/electron-current`.
+
+Note: If you're using `--xdg-layout`, then setting `npm_config_nodedir` should
```
gasinvein, line 224: `--xdg-layout` is enabled by default since #289

threema-danilo: @gasinvein I'm heading for vacations soon and don't have time to update this PR (which has been open and approved since March), but you should have permissions to push to my PR branch. Feel free to take over and push the desired changes directly to my fork!

gasinvein: Ok, edited.

threema-danilo: Thanks @gasinvein!

PR #263, opened 2022-04-06
Add --yaml flag to flatpak-pip-generator

JakobDev:
The new YAML flags allows writing the output as YAML instead of JSON.

### DIFF
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -13,6 +13,7 @@ import tempfile
 import urllib.request
 
 from collections import OrderedDict
+from typing import Dict
 
 try:
     import requirements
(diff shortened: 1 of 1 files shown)


TingPing:
I fail to see the value of this.

- YAML files can include JSON files
- These are automatically generated files that people will rarely read and shouldn't edit

JakobDev:
There are people who read it and I see some people who convert the generated JSON to YAML. This are only a few lines for people who prefer to have everything as YAML.

hfiguiere:
> There are people who read it 

sure, and JSON can be read too.

> and I see some people who convert the generated JSON to YAML. 

That doesn't justify it.

JakobDev:
> sure, and JSON can be read too.

It' harder to read than YAML

> That doesn't justify it.

If it actually used out there, there are a reason to implement this.

TingPing:
It is a really tiny patch so I won't argue against it too much.

PR #264, opened 2022-04-08
pip: expanduser on the requirements file argument

muelli:
Now you can run the script with --requirements-file=~/foo/requirements.txt.

### DIFF
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -139,12 +139,16 @@ def fprint(string: str) -> None:
 
 
 packages = []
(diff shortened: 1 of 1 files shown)


pip/flatpak-pip-generator
```diff
@@ -139,12 +139,12 @@ def fprint(string: str) -> None:
 
 
 packages = []
-if opts.requirements_file and os.path.exists(opts.requirements_file):
-    with open(opts.requirements_file, 'r') as req_file:
+if opts.requirements_file and os.path.exists(os.path.expanduser(opts.requirements_file)):
```
TingPing, line 142: To avoid repeating ourselves, just store this in an `expanded_requirements_file` variable and use that throughout.

PR #266, opened 2022-04-14
node: Option to get electron binaries per headers

gasinvein:
Electron version in `.yarnrc`/`.npmrc` may be different from the version in lockfile. In such case, if the version from rcfile is used for packaging, build will fail, since we only have the headers, but not the binaries for matching version.
Such case was observed with VSCode; despite it seems like a human error that is fixed by now, I believe `flatpak-node-generator` should be able to handle it.

This patch adds a cli option that tells the generator to download electron binaries for each electron version discovered in rcfiles.

### DIFF
--- a/node/flatpak-node-generator.py
+++ b/node/flatpak-node-generator.py
@@ -957,8 +957,8 @@ class SpecialSourceProvider:
             self.gen.add_url_source(integrity_file.url, integrity_file.integrity,
                                     electron_cache_dir / integrity_file.filename)
 
-    async def _handle_electron(self, package: Package) -> None:
-        manager = await ElectronBinaryManager.for_version(package.version)
+    async def __add_electron(self, version: str) -> None:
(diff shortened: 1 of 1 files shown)


node/flatpak-node-generator.py
```diff
@@ -1877,6 +1882,9 @@ async def main() -> None:
     parser.add_argument('--electron-node-headers',
                         action='store_true',
                         help='Download the electron node headers')
+    parser.add_argument('--electron-from-rcfile',
+                        action='store_true',
+                        help='Download electron version per node headers from .yarnrc/.npmrc')
```
refi64, line 1887: nit: `electron version corresponding to the node headers...`? I had to do a double-take with the string as written

gasinvein: Is it better now?

refi64: yup!

PR #268, opened 2022-04-19
pip: Add Message for PyQt BaseApp

JakobDev:
PyQt is widely used, but it is not supported by flatpak-pip-generator. There is now a [BaseApp for PyQt](https://github.com/flathub/com.riverbankcomputing.PyQt.BaseApp) for PyQt available. Instead of failing and leaving the User alone, flatpak-pip-generator will now provide a Link to the new BaseApp.

I know, that flatpak-pip-generator can also be used for other Repos than Flathub, but I think most of the Users are using it for Flathub. The others can go the the linked repo and copy the build files.

### DIFF
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -166,6 +166,13 @@ elif opts.packages:
 else:
     exit('Please specifiy either packages or requirements file argument')
 
+for i in packages:
+    if i["name"].lower().startswith("pyqt"):
+        print("PyQt packages are not supported by flapak-pip-generator")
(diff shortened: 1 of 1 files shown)


MehdiChinoune:
I use flatpak-pip-generator for flathub, but PyQt is not the base of my app. I use it to generate python binding for my application using sip. I am not alone I saw many projects do the same thing.

MehdiChinoune:
Also `pyqt-builder`.

TingPing:
@JakobDev Could you open a PR changing it to an explicit list rather than `pyqt-*`?

I'm not super familiar with pyqt personally.

PR #270, opened 2022-04-28
Add GitHub issue form

gasinvein:
See a preview [here](https://github.com/gasinvein/flatpak-builder-tools/issues/new?template=bug_template.yml).


gasinvein:
I was thinking about this, too, but didn't found a seamless way to do so. As the last resort, we probably can add an action that would apply  labels based on form data.

gasinvein:
It would be ridiculous amount of copy-paste. All the tools need essentially the same information in a bug report.

TingPing:
I don't believe thats a problem. It could be a preprocessed file or do some YAML magic to reduce duplication.

Issue #273, opened 2022-04-30
perl-libs/install.sh has the wrong prefix


Mailaender:
### flatpak-builder version

1.0.9

### Linux distribution and version

openSUSE 15.3

### Affected flatpak-builder tool

cpan/flatpak-cpan-generator.pl

### Description

`./flatpak-cpan-generator.pl Modern::Perl`

```yaml
modules:
  - name: perl-libs
    buildsystem: simple
    build-commands:
      - 'perl-libs/install.sh'
    # Same as with the Perl module, we need to restore write permission.
    # However, -f is now passed to avoid errors from trying to touch files from the
    # above module that are now marked as read-only.
    post-install:
      - 'chmod -Rf u+w /app/lib/perl5/site_perl'
    sources:
      - generated-sources.json
    # This step should be customized based on the CPAN packages you're using.
    cleanup:
      - '/bin'
      - '/man'
```

results in 

```
Running: perl-libs/install.sh
Checking if your kit is complete...
Looks good
Generating a Unix-style Makefile
Writing Makefile for Modern::Perl
Writing MYMETA.yml and MYMETA.json
cp lib/odern/Perl.pm blib/lib/odern/Perl.pm
cp lib/Modern/Perl.pm blib/lib/Modern/Perl.pm
Manifying 2 pod documents
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
ERROR: Can't create '/usr/lib/perl5/site_perl/5.34.0/Modern'
mkdir /usr/lib/perl5/site_perl/5.34.0/Modern: Read-only file system at /usr/lib/perl5/5.34.0/ExtUtils/Install.pm line 470.

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 at -e line 1.
make: *** [Makefile:732: pure_site_install] Error 30
```

refi64:
The runtime doesn't have Perl, so if you want to use it in your application, you need to add it to your manifest yourself. When you do so, its prefix will be properly set to `/app`, so any CPAN modules you install afterwards should go to the correct locations.

Mailaender:
I just needed a Perl script at build time and that already failed.

PR #274, opened 2022-05-01
Fixed prefix and destination directory for CPAN

Mailaender:
Closes https://github.com/flatpak/flatpak-builder-tools/issues/273

### DIFF
--- a/cpan/flatpak-cpan-generator.pl
+++ b/cpan/flatpak-cpan-generator.pl
@@ -122,7 +122,7 @@ sub main {
       "    mod_dir=\$1",
       "    cd \$mod_dir",
       "    if [ -f 'Makefile.PL' ]; then",
-      "        perl Makefile.PL && make install",
+      "        perl Makefile.PL PREFIX=${FLATPAK_DEST} && make install PREFIX=${FLATPAK_DEST}",
       "    elif [ -f 'Build.PL' ]; then",
       "        perl Build.PL && ./Build && ./Build install",
       "    else",


cpan/flatpak-cpan-generator.pl
```diff
@@ -122,7 +122,7 @@ sub main {
       "    mod_dir=\$1",
       "    cd \$mod_dir",
       "    if [ -f 'Makefile.PL' ]; then",
-      "        perl Makefile.PL && make install",
+      "        perl Makefile.PL PREFIX=/ && make install DESTDIR=/app",
```
TingPing, line 125: This sounds wrong. PREFIX would be `/app`, DESTDIR should be unset.

TingPing: Technically PREFIX is `/usr` if its a runtime and `/app` for applications, this script can, although probably isn't, usable for both.

Mailaender: It is counter intuitive, but this is what I get:

> Installing /app/lib/perl5/site_perl/5.34.0/x86_64-linux/auto/Compress/Raw/Lzma/Lzma.so
Installing /app/lib/perl5/site_perl/5.34.0/x86_64-linux/auto/Compress/Raw/Lzma/autosplit.ix
Installing /app/lib/perl5/site_perl/5.34.0/x86_64-linux/Compress/Raw/Lzma.pm
Installing /app/share/man/man3/Compress::Raw::Lzma.3
Appending installation info to /app//lib/perl5/5.34.0/x86_64-linux/perllocal.pod

Working test case https://github.com/Mailaender/flathub/tree/010editor

TingPing: I'm not a perl person but here is how these variables work in everything unix I've ever used.

- PREFIX is the actual prefix of the runtime environment, it is often passed to actual source code and is very relevant.
- DESTDIR is a prefix to copy files into, it is used only at install time and the purpose of it is for intermediate installation steps before final packaging. This does not apply to flatpak.

So while its possible Perl is doing something very odd. This just looks incorrect and risks the wrong PREFIX being used.

Mailaender: What do you suggest instead to write into the correct directories?

hfiguiere: `PREFIX=${FLATPAK_DEST}`

keep `DESTDIR` unset

TingPing: Are you setting prefix during both build and install?

Mailaender: That's it! Confirmed it works.

PR #281, opened 2022-05-15
cargo: Find Cargo.toml in repo subdirectories

### LINKED ISSUE #279: flatpak-cargo-generator.py fails when Cargo.toml is in subdirectory
### flatpak-builder version

1.2.2

Fedora 36

cargo/flatpak-cargo-generator.py

_No response_

### Source repository URL

https://github.com/lapce/lapce

_No response_

### Description

I'm currently building a flatpak for [Lapce](https://github.com/lapce/lapce). The flatpak already builds fine locally, but I also want to publish it on Flathub, so I need `flatpak-cargo-generator.py`.

However, running `flatpak-cargo-generator.py` with the lapce [v0.1.0 Cargo.lock file](https://raw.githubusercontent.com/lapce/lapce/v0.1.0/Cargo.lock) gives the following error:
```
  File "/home/user/src/lapce-flatpak/sourcebuild/./flatpak-cargo-generator.py", line 335, in <module>
    main()
  File "/home/user/src/lapce-flatpak/sourcebuild/./flatpak-cargo-generator.py", line 328, in main
    generated_sources = asyncio.run(generate_sources(load_toml(args.cargo_lock),
  File "/usr/lib64/python3.10/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/usr/lib64/python3.10/asyncio/base_events.py", line 646, in run_until_complete
    return future.result()
  File "/home/user/src/lapce-flatpak/sourcebuild/./flatpak-cargo-generator.py", line 282, in generate_sources
    for pkg in await asyncio.gather(*pkg_coros):
  File "/home/user/src/lapce-flatpak/sourcebuild/./flatpak-cargo-generator.py", line 235, in get_package_sources
    return await get_git_package_sources(package, git_repos)
  File "/home/user/src/lapce-flatpak/sourcebuild/./flatpak-cargo-generator.py", line 182, in get_git_package_sources
    git_repo['commits'][commit] = await get_git_repo_packages(repo_url, commit)
  File "/home/user/src/lapce-flatpak/sourcebuild/./flatpak-cargo-generator.py", line 106, in get_git_repo_packages
    root_toml = load_toml(os.path.join(git_repo_dir, 'Cargo.toml'))
  File "/home/user/src/lapce-flatpak/sourcebuild/./flatpak-cargo-generator.py", line 78, in load_toml
    with open(tomlfile, 'r') as f:
FileNotFoundError: [Errno 2] No such file or directory: '/home/user/.cache/flatpak-cargo/https_github.com_lapce_xi-editor/Cargo.toml'
```

The problem seems to be this entry:
```
[[package]]
name = "xi-rope"
version = "0.3.0"
source = "git+https://github.com/lapce/xi-editor#5578c7253ff8f644638bebace1261b6a412ce03c"
dependencies = [
 "bytecount",
 "memchr",
 "regex",
 "serde 1.0.130",
 "unicode-segmentation",
]
```
The source repository has a Cargo.toml file, but in a subdirectory called _rust_. Cargo seems to figure this out somehow, but `flatpak-cargo-generator.py` doesn't.

I tried to modify the following line to also check the _rust_ subdir:
https://github.com/flatpak/flatpak-builder-tools/blob/d20cdd501bba3c49adc9840004ce984740d1aeed/cargo/flatpak-cargo-generator.py#L106
This works, but then the dependencies of that package fail.
Does anyone know how Cargo handles this?

sebastian-de:
Some git repositories listed in Cargo.lock can have their Cargo.toml in a subdirectory instead of the root directory.
With this change, subdirectories get scanned for Cargo.toml as well, if it isn't found in the repository root.

I tested the change with a few Cargo.lock files and the output is the same for files that already worked before.

Fixes #279

### DIFF
--- a/cargo/flatpak-cargo-generator.py
+++ b/cargo/flatpak-cargo-generator.py
@@ -74,6 +74,27 @@ async def get_remote_sha256(url):
     return sha256.hexdigest()
 
 
+def find_cargo_toml(git_repo_dir):
+    if os.path.isfile(os.path.join(git_repo_dir, 'Cargo.toml')):
+        return load_toml(os.path.join(git_repo_dir, 'Cargo.toml')), '.'
+    # If Cargo.toml isn't found in repository root, scan subdirectories
+    toml_list = []
+    for entry in os.scandir(git_repo_dir):
+        if entry.is_dir():
+            toml_path = os.path.join(entry.path, 'Cargo.toml')
+            if os.path.isfile(toml_path):
+                toml_list.append((toml_path, os.path.relpath(entry.path, git_repo_dir)))
+    if len(toml_list) == 1:
+        return load_toml(toml_list[0][0]), toml_list[0][1]
+    if len(toml_list) > 1:
+        raise NotImplementedError(
+            f'Multiple Cargo.toml files found in {git_repo_dir}\n'
+            'Please report this error and the link to the affected repisitory here:\n'
+            'https://github.com/flatpak/flatpak-builder-tools/issues'
+            )
+    raise Exception(f'No Cargo.toml found in {git_repo_dir}')
+
+
 def load_toml(tomlfile='Cargo.lock'):
     with open(tomlfile, 'r') as f:
         toml_data = toml.load(f)
@@ -103,7 +124,7 @@ def fetch_git_repo(git_url, commit):
 async def get_git_repo_packages(git_url, commit):
     logging.info('Loading packages from %s', git_url)
     git_repo_dir = fetch_git_repo(git_url, commit)
-    root_toml = load_toml(os.path.join(git_repo_dir, 'Cargo.toml'))
+    root_toml, toml_dir = find_cargo_toml(git_repo_dir)
     assert 'package' in root_toml or 'workspace' in root_toml
     packages = {}
 
@@ -128,7 +149,7 @@ async def get_git_repo_packages(git_url, commit):
                 await get_dep_packages(target, toml_dir)
 
     if 'package' in root_toml:
-        await get_dep_packages(root_toml, '.')
+        await get_dep_packages(root_toml, toml_dir)
         packages[root_toml['package']['name']] = '.'
 
     if 'workspace' in root_toml:


hfiguiere:
where is this documented in Cargo?

sebastian-de:
I didn't find anything in the official documentation, only [this comment](https://github.com/rust-lang/cargo/issues/9011#issuecomment-754169946).

And a hint in the cargo source code: https://github.com/rust-lang/cargo/blob/2ff0b401b9c79a44789a71eb53d83d422bb9b200/src/cargo/ops/common_for_install_and_uninstall.rs#L617

hfiguiere:
ah ok.

cargo/flatpak-cargo-generator.py
```diff
@@ -103,7 +103,18 @@ def fetch_git_repo(git_url, commit):
...
+        root_toml = load_toml(os.path.join(git_repo_dir, 'Cargo.toml'))
+        toml_dir = '.'
+    else:
+        for entry in os.scandir(git_repo_dir):
+            if entry.is_dir():
+                toml_path = os.path.join(git_repo_dir, entry.name, 'Cargo.toml')
+                if os.path.isfile(toml_path):
+                    root_toml = load_toml(toml_path)
+                    toml_dir = entry.name
+                    break
```
gasinvein, line 116: Picking up the first found (mostly random) `Cargo.toml` file doesn't seem like the proper way to go.

And what we do if didn't find a `Cargo.toml`? `root_toml` becomes undefined in this case.

gasinvein (changes requested):

cargo/flatpak-cargo-generator.py
sebastian-de: The script now raises an exception if multiple or none Corgo.toml are found.

cargo/flatpak-cargo-generator.py
```diff
@@ -74,6 +74,27 @@ async def get_remote_sha256(url):
...
+    if os.path.isfile(os.path.join(git_repo_dir, 'Cargo.toml')):
+        toml_data = load_toml(os.path.join(git_repo_dir, 'Cargo.toml'))
+        toml_dir = '.'
+    else:
+        toml_list = []
+        for entry in os.scandir(git_repo_dir):
+            if entry.is_dir():
+                toml_path = os.path.join(git_repo_dir, entry.name, 'Cargo.toml')
+                if os.path.isfile(toml_path):
+                    toml_list.append((toml_path, entry.name))
```
gasinvein, line 87: Should be not just dir name but a path relative to git repo root, I suppose.

cargo/flatpak-cargo-generator.py
```diff
@@ -74,6 +74,27 @@ async def get_remote_sha256(url):
...
+        if len(toml_list) == 1:
+            toml_data = load_toml(toml_list[0][0])
+            toml_dir = toml_list[0][1]
+        elif len(toml_list) > 1:
+            raise Exception(f'Multiple Cargo.toml files found in {git_repo_dir}')
+        else:
+            raise Exception(f'No Cargo.toml found in {git_repo_dir}')
```
gasinvein, line 94: We potentially can handle repos with multiple `Cargo.toml`s in subdirectories and without top-level `Cargo.toml`, just not yet. Maybe raise a `NotImplementedError` a suggest the user to file an issue pointing to the the unsupported repo?

cargo/flatpak-cargo-generator.py
```diff
@@ -74,6 +74,27 @@ async def get_remote_sha256(url):
     return sha256.hexdigest()
 
 
+def find_toml(git_repo_dir):
+    if os.path.isfile(os.path.join(git_repo_dir, 'Cargo.toml')):
+        toml_data = load_toml(os.path.join(git_repo_dir, 'Cargo.toml'))
+        toml_dir = '.'
```
gasinvein, line 80: We can just `return` the values here, then the `else` clause won't be needed.

cargo/flatpak-cargo-generator.py
```diff
@@ -74,6 +74,27 @@ async def get_remote_sha256(url):
...
+    else:
+        toml_list = []
+        for entry in os.scandir(git_repo_dir):
+            if entry.is_dir():
+                toml_path = os.path.join(git_repo_dir, entry.name, 'Cargo.toml')
+                if os.path.isfile(toml_path):
+                    toml_list.append((toml_path, entry.name))
+        if len(toml_list) == 1:
+            toml_data = load_toml(toml_list[0][0])
+            toml_dir = toml_list[0][1]
```
gasinvein, line 90: Same as above, we can just `return` the values.

gasinvein (changes requested):

cargo/flatpak-cargo-generator.py
```diff
@@ -74,6 +74,28 @@ async def get_remote_sha256(url):
...
+                toml_list.append((toml_path, os.path.relpath(entry.path, git_repo_dir)))
+    if len(toml_list) == 1:
+        return load_toml(toml_list[0][0]), toml_list[0][1]
+    if len(toml_list) > 1:
+        raise NotImplementedError(
+            f'Multiple Cargo.toml files found in {git_repo_dir}\n'
+            'Please report this error and the link to the affected repisitory here:\n'
+            'https://github.com/flatpak/flatpak-builder-tools/issues'
+            )
+    if not toml_list:
```
gasinvein, line 95: This `if` clause is redundant, since we have either `raise`d or `return`ed before it's reached.

gasinvein (changes requested):

cargo/flatpak-cargo-generator.py
```diff
@@ -74,6 +74,28 @@ async def get_remote_sha256(url):
...
 
 
+def find_toml(git_repo_dir):
+    if os.path.isfile(os.path.join(git_repo_dir, 'Cargo.toml')):
+        return load_toml(os.path.join(git_repo_dir, 'Cargo.toml')), '.'
+    # If Cargo.toml isn't found in repository root, scan subdirectories
+    toml_list = []
+    for entry in os.scandir(git_repo_dir):
+        if entry.is_dir():
+            toml_path = os.path.join(git_repo_dir, entry.name, 'Cargo.toml')
```
gasinvein, line 84: I guess this will work, too?
```suggestion
            toml_path = os.path.join(entry.path, 'Cargo.toml')
```

cargo/flatpak-cargo-generator.py
sebastian-de: `len(toml_list)` can still be `0`, that's what that line is for.
I can write it as `len(toml_list) == 0` to make that clear.

gasinvein: What I'm saying is that `len(toml_list)` can be *only* equal to zero at this point, since we've already checked if it's exactly 1 or more than 1. So, there is no need to check for equality to zero explicitly.

gasinvein (approved):
LGTM now, thanks. I presume it still works for your case?

PR #282, opened 2022-05-26
* Fix support for Electron v19+.

MrMYHuang:
From Electron release page, it can be seen that v19 and v20 drop the linux-ia32 supports:
https://github.com/electron/electron/releases

This commit skips linux-ia32 for Electron v19+.

### DIFF
--- a/node/flatpak-node-generator.py
+++ b/node/flatpak-node-generator.py
@@ -851,6 +851,10 @@ class ElectronBinaryManager:
 
     def find_binaries(self, binary: str) -> Iterator['ElectronBinaryManager.Binary']:
         for electron_arch, flatpak_arch in self.ELECTRON_ARCHES_TO_FLATPAK.items():
+            # Electron v19+ drop linux-ia32 support.
+            if SemVer.parse(self.version) >= SemVer.parse("19.0.0") and electron_arch == "ia32":
+                continue
+
             binary_filename = f'{binary}-v{self.version}-linux-{electron_arch}.zip'
             binary_url = self.child_url(binary_filename)
 


hfiguiere:
i386 hasn't been supported on flathub in like 3 years so no big deal :-D

MrMYHuang:
> i386 hasn't been supported on flathub in like 3 years so no big deal :-D

So, is the better solution to remove ia32 support from flatpak-node-generator.py?
(Without fixing this issue, flatpak-node-generator.py will run to fail for Electron v19+.)

hfiguiere:
I don't think it's necessary to rip it out (however I don't maintain this, so ultimately it's not my call). I was just mentionning that in the case of flathub it was basically a no-op.

MrMYHuang:
> I don't think it's necessary to rip it out (however I don't maintain this, so ultimately it's not my call). I was just mentionning that in the case of flathub it was basically a no-op.

Yes. flatpak-builder-tools can be used not only for Flathub, but also e.g., a self-hosted flatpak repo.

node/flatpak-node-generator.py
```diff
@@ -851,6 +851,10 @@ def child_url(self, child: str) -> str:
 
     def find_binaries(self, binary: str) -> Iterator['ElectronBinaryManager.Binary']:
         for electron_arch, flatpak_arch in self.ELECTRON_ARCHES_TO_FLATPAK.items():
+            # Electron v19+ drop linux-ia32 support.
+            if version_tuple(self.version) >= version_tuple("19.0.0") and electron_arch == "ia32":
```
gasinvein, line 855: We already have a custom version parser.
```suggestion
            if SemVer.parse(self.version) >= SemVer.parse("19.0.0") and electron_arch == "ia32":
```

MrMYHuang: Done. Please check [b60bc6f](https://github.com/flatpak/flatpak-builder-tools/pull/282/commits/b60bc6f0fcc6f761c979c558726ffd53f20afbd3). Thanks.

gasinvein (approved):
LGTM, thanks.

PR #287, opened 2022-06-03
node: Get registry integrity matching lockfile

gasinvein:
The problem:
To my understanding, if a package in the lockfile has `sha1` integrity, `npm` will look for `sha1` metadata in cache, but we only have `sha512` there since it's what the registry usually provides - and we discard the lockfile integrity and use the registry one.

The proposed solution:
Keep the lockfile integrity value; when fetching integrity from the registry, make sure it matches the one that was in the lockfile.

Fixes the `npm install` error in flathub/chat.delta.desktop#93

### DIFF
--- a/node/flatpak-node-generator.py
+++ b/node/flatpak-node-generator.py
@@ -561,10 +561,6 @@ class ResolvedSource(NamedTuple):
             return metadata.integrity
 
 
-class UnresolvedRegistrySource:
-    pass
-
-
 class GitSource(NamedTuple):
     original: str
     url: str
@@ -572,7 +568,7 @@ class GitSource(NamedTuple):
     from_: Optional[str]
 
 
-PackageSource = Union[ResolvedSource, UnresolvedRegistrySource, GitSource]
+PackageSource = Union[ResolvedSource, GitSource]
 
 
 class Package(NamedTuple):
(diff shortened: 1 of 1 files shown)


refi64:
Ack, I'm moderately sure using the other checksum specifically was intentional *at some point*...could it be behavior that changed across npm versions, or maybe a side effect of the lockfile v2 transition?

gasinvein:
Wait a moment... Previous comment said:

> npm ignores the resolved field and just uses the provided registry instead.

How was this even possible? If NPM is running offline, it can't know what kind of digest is in the registry. Thus it can rely only on the `integrity` field in the lockfile. Or am I missing something?

refi64:
Hahaha, so after hitting this locally, I've come to believe this *actually changed in the npm registry*, and dists that formerly returned the sha1 they were uploaded with now return a newly computed sha512. I think the previous behavior I followed was also [changed in npm/pacote at some point](https://github.com/npm/pacote/commit/e8c274c85991b74ca0bb7cc5c68be647cf5cbbfd) too.

In addition, I found this mysterious comment in the pacote source code:

```
    // XXX pacote <=9 has some logic to ignore opts.resolved if
    // the resolved URL doesn't go to the same registry.
    // Consider reproducing that here, to throw away this.resolved
    // in that case.
```

...which I'm guessing is somehow related to why I was ignoring `resolved` altogether before. I'll redo this change and drop it on top of my current, refactored fork and post it SOON(tm) (probably tomorrow).

gasinvein:
Superseded by #289

PR #289, opened 2022-06-16
Refactor the node generator into a proper Python package & add tests

refi64:
...and also fix some bugs (includes slightly-tweaked #287). But bugs don't matter anyway, because THE CODE IS PRETTY!

Rationale:
- Making a Python package: Once published, it'll be easily installable w/ `pipx install flatpak-node-generator`, and *we can actually safely add dependencies* without breaking everything or requiring manual venv management.
  - but I did keep our own semver for now, because apparently the packaging module's doesn't actually support some version values that actual semver does.
- Type checking: This thing is *massive*, type checking helps keep things functioning.
- Formatting: Okay this is totally just that I really like pretty code, and I also refuse to let go of my single quotes.

The main still-outstanding things are:

- More tests! So many more tests!
- I need to add GitHub Actions support to auto-run the tests on changes.
- Later on, I want to fix up the console output a bit (did I mention we can safely add dependencies now?).

But for now, this is a good start!

(If I sound slightly hyper after spending several hours fighting with npm, it is because I am indeed hyper after spending several hours fighting with npm.)


node/webpack-quick-start/build.electron.webpack.ElectronWebpackQuickStart.yaml
```diff
@@ -1,84 +1,83 @@
...
+        # Set the cache directory, used to find Electron and some other helper
+        # tools.
+        # (The directory format is: /run/build/MODULE_NAME/flatpak-node/cache)
+        XDG_CACHE_HOME: /run/build/electron-webpack-quick-start/flatpak-node/cache
+        # Sets the directory where Node is located so that npm won't download the headers.
+        npm_config_nodedir: /usr/lib/sdk/node12
     build-commands:
       # Have Yarn use the offline mirror.
-      - 'HOME=$PWD yarn config --offline set yarn-offline-mirror $FLATPAK_BUILDER_BUILDDIR/flatpak-node/yarn-mirror'
+      - HOME=$PWD yarn config --offline set yarn-offline-mirror $FLATPAK_BUILDER_BUILDDIR/flatpak-node/yarn-mirror
```
gasinvein, line 45: This is rather hackish, I'd rather suggest adding a separate `.yarnrc` source instead:
```yaml
sources:
  ...
  - type: inline
    contents: |
      yarn-offline-mirror /run/build/electron-webpack-quick-start/flatpak-node/yarn-mirror
    dest-filename: .yarnrc

PR #300, opened 2022-07-05
node: Allow empty `resolved` field in lockfiles

### LINKED ISSUE #295: node: Fails to read `package-lock.json` entries without `resolved` field
### flatpak-builder version

1.2.2

Fedora 36

-r

### Source repository URL

https://github.com/bitwarden/desktop.git

_No response_

### Description

Seems like a regression introduced in #287, merged as part of #289
Quickfix is as simple as making the `ResolvedSource.resolved` property optional:
```patch
diff --git a/node/flatpak_node_generator/providers/npm.py b/node/flatpak_node_generator/providers/npm.py
index eeab9cf..e991884 100644
--- a/node/flatpak_node_generator/providers/npm.py
+++ b/node/flatpak_node_generator/providers/npm.py
@@ -65,7 +65,7 @@ class NpmLockfileProvider(LockfileProvider):
                 source = LocalSource(path=version[len('file:'):])
             else:
                 integrity = Integrity.parse(info['integrity'])
-                source = ResolvedSource(resolved=info['resolved'], integrity=integrity)
+                source = ResolvedSource(resolved=info.get('resolved'), integrity=integrity)
 
             yield Package(name=name, version=version, source=source, lockfile=lockfile)
 
```
although I'm not sure if it's a proper fix.

gasinvein:
We don't use the `resolved` value anywhere in the npm codepath, so it seems safe to make the field optional.
Renamed `ResolvedSource` to `RegistrySource` because objects of the former class without an actual `resolved` value don't feel right.

Fixes #295
Supersedes #241

### DIFF
--- a/node/flatpak_node_generator/providers/npm.py
+++ b/node/flatpak_node_generator/providers/npm.py
@@ -34,6 +34,10 @@ from ..url_metadata import RemoteUrlMetadata
 from . import LockfileProvider, ModuleProvider, ProviderFactory, RCFileProvider
 from .special import SpecialSourceProvider
 
+_NPM_CORGIDOC = (
+    'application/vnd.npm.install-v1+json; q=1.0, application/json; q=0.8, */*'
+)
+
 
 class NpmLockfileProvider(LockfileProvider):
     _ALIAS_RE = re.compile(r'^npm:(.[^@]*)@(.*)$')
@@ -163,8 +167,14 @@ class NpmModuleProvider(ModuleProvider):
             / self.get_cacache_integrity_path(integrity)
         )
 
-    def add_index_entry(self, url: str, metadata: RemoteUrlMetadata) -> None:
+    def add_index_entry(
+        self,
+        url: str,
+        metadata: RemoteUrlMetadata,
+        request_headers: Dict[str, str] = {},
+    ) -> None:
         key = f'make-fetch-happen:request-cache:{url}'
+
         index_json = json.dumps(
             {
                 'key': key,
@@ -173,7 +183,7 @@ class NpmModuleProvider(ModuleProvider):
                 'size': metadata.size,
                 'metadata': {
                     'url': url,
-                    'reqHeaders': {},
+                    'reqHeaders': request_headers,
                     'resHeaders': {},
                 },
             }
@@ -304,7 +314,9 @@ class NpmModuleProvider(ModuleProvider):
             )
             content_path = self.get_cacache_content_path(metadata.integrity)
             self.gen.add_data_source(raw_data, content_path)
-            self.add_index_entry(index.url, metadata)
+            self.add_index_entry(
+                index.url, metadata, request_headers={'accept': _NPM_CORGIDOC}
+            )
 
         patch_commands: DefaultDict[Path, List[str]] = collections.defaultdict(
             lambda: []
--- a/node/flatpak_node_generator/providers/npm.py
+++ b/node/flatpak_node_generator/providers/npm.py
@@ -21,7 +21,14 @@ import types
 
 from ..integrity import Integrity
 from ..manifest import ManifestGenerator
-from ..package import GitSource, LocalSource, Package, PackageSource, ResolvedSource
+from ..package import (
+    GitSource,
+    LocalSource,
+    Package,
+    PackageSource,
+    RegistrySource,
+    ResolvedSource,
+)
 from ..requests import Requests
 from ..url_metadata import RemoteUrlMetadata
 from . import LockfileProvider, ModuleProvider, ProviderFactory, RCFileProvider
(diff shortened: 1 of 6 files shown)


node/flatpak_node_generator/providers/npm.py
```diff
@@ -64,8 +64,10 @@ def process_dependencies(
             elif version.startswith('file:'):
                 source = LocalSource(path=version[len('file:') :])
             else:
-                integrity = Integrity.parse(info['integrity'])
-                source = ResolvedSource(resolved=info['resolved'], integrity=integrity)
+                source = RegistrySource(
+                    resolved=info.get('resolved'),
```
refi64, line 68: ack, this is making me want to bring back `UnresolvedRegistrySource` which was like, [just removed](https://github.com/flatpak/flatpak-builder-tools/pull/289/commits/7fc0cf2ecf7f20974d79d08a3d51fab1954bd826) :sweat_smile: 

Since right now, ResolvedSource/RegistrySource is used to hold two different things:

1. an unresolved package, with only an integrity to go off of
2. a fully resolved package

In npm-land, a source transitions from 1->2 with `resolve_source`, after which we know it's resolved. In yarn-land, it's *always* resolved...but in both cases, now we need to handle the case where `resolved is None`.

So let's hypothetically say `UnresolvedRegistrySource` is brought back, but now with a single member `integrity: Integrity`. and `resolve_source` takes a Package with an `UnresolvedRegistrySource` inside (just like before). Then, *all* of the assertions on `resolved` could actually be removed, because it's virtually guaranteed our source values are always resolved for us already.

(Of course really, the assertions etc are arguably off-topic for this PR anyway, but at minimum I feel like bringing back `UnresolvedRegistrySource` for this purpose is a bit nicer anyway!)

(also ngl I feel like the original intent of `UnresolvedRegistrySource` in the first place was something like this? but it's been so long I forgot the details...)

refi64:
#295 says this is from Bitwarden, is it from the normal master branch there? I was trying to see which package was missing the field, but I can't reproduce the issue at all... (unless poetry is just running the old node-generator version for some reason, but I want to double-check before I delve into venv weirdness :sweat_smile: )

node/flatpak_node_generator/providers/npm.py
gasinvein: Maybe make make `*Source` classes datacalsses instead of NamedTuples, then? If bring `UnresolvedRegistrySource` back, it feels natural to use inheritance here, like
```python
class RegistrySource(PackageSource):
    pass

class UnresolvedRegistrySource(RegistrySource):
    integrity: Integrity

class ResolvedRegistrySource(RegistrySource):
    resolved: str
    integrity: Integrity
```
or even
```python
class UnresolvedRegistrySource(PackageSource):
    integrity: Integrity

class ResolvedRegistrySource(UnresolvedRegistrySource):
    resolved: str
```

catsout:
https://github.com/laurent22/joplin/issues/5658  

The `resolved` filed may be empty when a lockfile is damaged, or very old.  
If we handle this, we also need https://github.com/flatpak/flatpak-builder-tools/pull/241, as when `resolved`  is empty, npm will try to query non tgz index_entry.

refi64:
ack, thanks for the reminder of that PR! I did some digging into Joplin and [created a test case based on what I found](https://paste.sr.ht/~refi64/eeb91c56e3a6272070c822552d744f32e1852aa4#add-tests.patch), and it *does* indeed fail to build. If I re-apply/-base the fix from #241 ([with some stylistic changes](https://paste.sr.ht/~refi64/eeb91c56e3a6272070c822552d744f32e1852aa4#corgidoc.patch)), then it does, in fact, work. With that in mind, this PR would also need that change + the test case. @catsout do you want to try to PR the updated changes to @gasinvein's fork so it ends up in this PR, or I can send it but credit you as the author?

node/flatpak_node_generator/providers/npm.py
refi64: I was actually going to try sending a PR to switch these to dataclasses after this was merged, was planning to before and then I forgot. With that in mind, I'd definitely be fine with changing it here first. (the latter class layout is definitely nicer than what I had before!)

catsout:
@refi64 
You can send it but credit me as the author.

node/flatpak_node_generator/providers/npm.py
gasinvein: Ok, done.
It seems natural now to switch other similar things (like `Package`) to dataclasses, but it's out of scope of this PR, so I didn't touch them.

gasinvein:
Thanks, I've applied both the @refi64's test case and redone #241. @catsout, can you confirm it works for Joplin?

catsout:
@gasinvein 
Yes, I can. It may be not enough to build joplin entirely(need some git patches), but should fix the empty `resolved` problem.

PR #303, opened 2022-07-10
node: Support url as dependency

### LINKED ISSUE #145: node: AssertionError on package version although it is available
I tried generating the sources from the following [lock file](https://github.com/zulip/zulip-desktop/blob/4b3608fc1ee58793560135ba53a00301592d4877/package-lock.json).

I am getting the following assertion errors:

```
Found 1 lockfiles.
Reading packages from lockfiles...
1475 packages read.
Generating packages [1473/1475] ms @ 2.0.0Task exception was never retrieved                             
future: <Task finished coro=<GeneratorProgress._generate() done, defined at ./node/flatpak-node-generator.py:1431> exception=AssertionError('entities versions available are 0.1.0, 0.1.1, 0.2.0, 0.2.1, 0.2.2, 0.3.0, 0.4.0, 0.4.1, 0.5.0, 1.0.0, 1.1.0, 1.1.1, 1.1.2, 2.0.0, 2.0.1, 2.0.2, 2.0.3, not https://registry.npmjs.org/entities/-/entities-1.1.2.tgz')>
  File "./node/flatpak-node-generator.py", line 1433, in _generate
    await self.module_provider.generate_package(package)
  File "./node/flatpak-node-generator.py", line 1110, in generate_package
    source = await self.resolve_source(package)
  File "./node/flatpak-node-generator.py", line 1086, in resolve_source
    f'{package.name} versions available are {", ".join(versions)}, not {package.version}'
AssertionError: entities versions available are 0.1.0, 0.1.1, 0.2.0, 0.2.1, 0.2.2, 0.3.0, 0.4.0, 0.4.1, 0.5.0, 1.0.0, 1.1.0, 1.1.1, 1.1.2, 2.0.0, 2.0.1, 2.0.2, 2.0.3, not https://registry.npmjs.org/entities/-/entities-1.1.2.tgz
Task exception was never retrieved
future: <Task finished coro=<GeneratorProgress._generate() done, defined at ./node/flatpak-node-generator.py:1431> exception=AssertionError('readable-stream versions available are 0.0.1, 0.0.2, 0.0.3, 0.0.4, 0.1.0, 0.2.0, 0.3.0, 0.3.1, 1.0.0, 1.0.1, 1.0.2, 1.0.15, 1.0.17, 1.1.7, 1.1.8, 1.1.9, 1.0.24, 1.0.25, 1.1.10, 1.0.25-1, 1.1.11, 1.0.26, 1.0.26-1, 1.1.11-1, 1.0.26-2, 1.1.12, 1.0.26-3, 1.0.26-4, 1.1.12-1, 1.0.27-1, 1.1.13-1, 1.0.31, 1.1.13, 1.0.32, 1.0.32-1, 1.0.33-1, 1.0.33-2, 1.0.33, 2.0.0, 2.0.1, 2.0.2, 2.0.3, 2.0.4, 2.0.5, 2.0.6, 1.0.34, 1.1.14, 2.1.0, 2.1.1, 2.1.2, 2.1.3, 2.1.4, 2.1.5, 2.2.0, 2.2.1, 2.2.2, 2.2.3, 2.2.4, 2.2.5, 2.2.6, 2.2.7, 2.2.8, 2.2.9, 2.2.10, 2.2.11, 2.3.0, 2.3.1, 2.3.2, 2.3.3, 2.3.4, 2.3.5, 2.3.6, 3.0.0-rc.1, 3.0.0-rc.2, 3.0.0-rc.3, 3.0.0, 3.0.1, 3.0.2, 3.0.3, 3.0.4, 3.0.5, 3.0.6, 3.1.0, 3.1.1, 3.2.0, 3.3.0, 3.4.0, 2.3.7, 3.5.0, 3.6.0, not https://registry.npmjs.org/readable-stream/-/readable-stream-3.6.0.tgz')>
  File "./node/flatpak-node-generator.py", line 1433, in _generate
    await self.module_provider.generate_package(package)
  File "./node/flatpak-node-generator.py", line 1110, in generate_package
    source = await self.resolve_source(package)
  File "./node/flatpak-node-generator.py", line 1086, in resolve_source
    f'{package.name} versions available are {", ".join(versions)}, not {package.version}'
AssertionError: readable-stream versions available are 0.0.1, 0.0.2, 0.0.3, 0.0.4, 0.1.0, 0.2.0, 0.3.0, 0.3.1, 1.0.0, 1.0.1, 1.0.2, 1.0.15, 1.0.17, 1.1.7, 1.1.8, 1.1.9, 1.0.24, 1.0.25, 1.1.10, 1.0.25-1, 1.1.11, 1.0.26, 1.0.26-1, 1.1.11-1, 1.0.26-2, 1.1.12, 1.0.26-3, 1.0.26-4, 1.1.12-1, 1.0.27-1, 1.1.13-1, 1.0.31, 1.1.13, 1.0.32, 1.0.32-1, 1.0.33-1, 1.0.33-2, 1.0.33, 2.0.0, 2.0.1, 2.0.2, 2.0.3, 2.0.4, 2.0.5, 2.0.6, 1.0.34, 1.1.14, 2.1.0, 2.1.1, 2.1.2, 2.1.3, 2.1.4, 2.1.5, 2.2.0, 2.2.1, 2.2.2, 2.2.3, 2.2.4, 2.2.5, 2.2.6, 2.2.7, 2.2.8, 2.2.9, 2.2.10, 2.2.11, 2.3.0, 2.3.1, 2.3.2, 2.3.3, 2.3.4, 2.3.5, 2.3.6, 3.0.0-rc.1, 3.0.0-rc.2, 3.0.0-rc.3, 3.0.0, 3.0.1, 3.0.2, 3.0.3, 3.0.4, 3.0.5, 3.0.6, 3.1.0, 3.1.1, 3.2.0, 3.3.0, 3.4.0, 2.3.7, 3.5.0, 3.6.0, not https://registry.npmjs.org/readable-stream/-/readable-stream-3.6.0.tgz
```

All packages seem to have their expected version available for download. The error message suggest some parsing error in the version, as I don't expect to see the full URL in the `package.version` field.

### LINKED ISSUE #301: node-generator fails because npm registry is the default
When a package have different registry node generator fails to generate sources because it looks for that package in npm.

Example:
* https://github.com/cinnyapp/cinny/blob/dev/package-lock.json#L2407  
* Example run https://github.com/kfiven/cinny-desktop/runs/7216739597?check_suite_focus=true#step:10:3095

gasinvein:
"[URL as dependency](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#urls-as-dependencies)" is the term used by NPM documentation; in fact, it refers to direct package URLs in the `version` field.

Should fix #145, and probably fix #301

### DIFF
--- a/node/flatpak_node_generator/package.py
+++ b/node/flatpak_node_generator/package.py
@@ -87,12 +87,12 @@ class PackageSource(abc.ABC):
 
 
 @dataclass(frozen=True, eq=True)
-class RegistrySource(PackageSource):
+class PackageFileSource(PackageSource):
     integrity: Optional[Integrity]
 
 
 @dataclass(frozen=True, eq=True)
-class ResolvedSource(RegistrySource):
+class PackageURLSource(PackageFileSource):
     resolved: str
 
     async def retrieve_integrity(self) -> Integrity:
@@ -103,6 +103,16 @@ class ResolvedSource(RegistrySource):
             assert url is not None, 'registry source has no resolved URL'
             metadata = await RemoteUrlMetadata.get(url, cachable=True)
             return metadata.integrity
+
+
+@dataclass(frozen=True, eq=True)
+class RegistrySource(PackageFileSource):
+    pass
+
+
+@dataclass(frozen=True, eq=True)
+class ResolvedSource(RegistrySource, PackageURLSource):
+    pass
 
 
 @dataclass(frozen=True, eq=True)
--- a/node/flatpak_node_generator/providers/npm.py
+++ b/node/flatpak_node_generator/providers/npm.py
@@ -27,6 +27,7 @@ from ..package import (
     LocalSource,
     Package,
     PackageSource,
+    PackageURLSource,
     RegistrySource,
     ResolvedSource,
 )
@@ -82,6 +83,8 @@ class NpmLockfileProvider(LockfileProvider):
                     source = ResolvedSource(
                         resolved=info['resolved'], integrity=integrity
                     )
+                elif version_url.scheme in {'http', 'https'}:
+                    source = PackageURLSource(resolved=version, integrity=integrity)
                 else:
                     source = RegistrySource(integrity=integrity)
 
@@ -280,6 +283,24 @@ class NpmModuleProvider(ModuleProvider):
             self.add_index_entry(source.resolved, metadata)
 
             await self.special_source_provider.generate_special_sources(package)
+
+        elif isinstance(source, PackageURLSource):
+            assert source.integrity is not None
+
+            self.gen.add_url_source(
+                url=source.resolved,
+                integrity=source.integrity,
+                destination=self.get_cacache_content_path(source.integrity),
+            )
+            self.add_index_entry(
+                url=source.resolved,
+                metadata=RemoteUrlMetadata(
+                    integrity=source.integrity,
+                    size=await RemoteUrlMetadata.get_size(
+                        source.resolved, cachable=True
+                    ),
+                ),
+            )
 
         # pyright: reportUnnecessaryIsInstance=false
         elif isinstance(source, GitSource):
--- a/node/flatpak_node_generator/providers/npm.py
+++ b/node/flatpak_node_generator/providers/npm.py
@@ -18,6 +18,7 @@ import re
 import shlex
 import textwrap
 import types
+import urllib.parse
 
 from ..integrity import Integrity
 from ..manifest import ManifestGenerator
@@ -59,6 +60,7 @@ class NpmLockfileProvider(LockfileProvider):
                 continue
 
             version: str = info['version']
+            version_url = urllib.parse.urlparse(version)
             alias_match = self._ALIAS_RE.match(version)
             if alias_match is not None:
                 name, version = alias_match.groups()
@@ -72,8 +74,8 @@ class NpmLockfileProvider(LockfileProvider):
                     from_ = from_[match.end('prefix') :]
 
                 source = self.parse_git_source(version, from_)
-            elif version.startswith('file:'):
-                source = LocalSource(path=version[len('file:') :])
+            elif version_url.scheme == 'file':
+                source = LocalSource(path=version_url.path)
             else:
                 integrity = Integrity.parse(info['integrity'])
                 if 'resolved' in info:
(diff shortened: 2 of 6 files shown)


kfiven:
About #301 It built the sources successfully https://github.com/kfiven/cinny-desktop/actions/runs/2644846386

node/flatpak_node_generator/providers/npm.py
```diff
@@ -19,13 +19,16 @@
 import textwrap
 import types
 
+from yarl import URL
```
refi64, line 22: Any reason we can't use urllib here? (If that's not workable it's fine too, but then this should be in the `pyproject.toml`.)

gasinvein: I just like `yarl` more for the pathlib-like `URL() / ""` syntax. But actually, it's not really beneficial here; will replace with `urllib`.

node/flatpak_node_generator/package.py
```diff
@@ -105,6 +105,16 @@ async def retrieve_integrity(self) -> Integrity:
             return metadata.integrity
 
 
+@dataclass(frozen=True, eq=True)
+class RegistrySource(PackageFileSource):
+    pass
+
+
+@dataclass(frozen=True, eq=True)
+class ResolvedSource(RegistrySource, PackageURLSource):
```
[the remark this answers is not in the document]
gasinvein: > Can we leave the class hierarchy as before & either just use `ResolvedSource` as-is, or rename it?

I've added a new class to indicate where the source originated from. "URL as dependency" sources don't originate from a registry from npm perspective, and we don't need to generate registry index caches for them, so I concluded they shouldn't be inherited from `ResolvedSource`.
In other words, `isinstance(PackageURLSource(), ResolvedSource) == False` is intentional.

node/flatpak_node_generator/providers/npm.py
refi64: (TBH if you have package suggestions in general, I'd be open to hear them, since `pipx` == actually safe installs now.)

node/flatpak_node_generator/package.py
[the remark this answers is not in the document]
refi64: Hmm...that makes sense, and these are pretty lightweight dataclasses anyway.

node/flatpak_node_generator/providers/npm.py
gasinvein: Done.

PR #309, opened 2022-08-04
node: Update ms-playwright support

gasinvein:


kanru:
Confirmed this is required to work with playwright-1.24.2

gasinvein:
I believe the revision number scheme changed after microsoft/playwright@67c251f4837c546b14bcb7779304992f5e5ac953; so, the first playwright version to use sequential revision numbers for chromium should be `1.21.0`.

node/flatpak_node_generator/providers/special.py
```diff
@@ -328,7 +328,12 @@ async def _handle_playwright(self, package: Package) -> None:
             revision = int(browser['revision'])
 
             if name == 'chromium':
-                if revision < 792639:
+                # Revision number scheme was changed from Chromium version to incrementing
+                # integers above 1000; now it's same as for other browsers
+                if (
+                    SemVer.parse(package.version) < SemVer.parse('1.21.0')
+                    and revision < 792639
```
refi64, line 335: small nit: I think 792639 is the Chromium *revision*, not just the version? Maybe it's just my tired brain, but I spent a very hot minute trying to figure out what on earth Chromium version 792639 is supposed to be :sweat_smile:

gasinvein: Right, fixed the comment.

PR #313, opened 2022-08-20
node: Skip local source check in node_modules dir

### LINKED ISSUE #312: AssertionError on npm local source unavailable before build
### flatpak-builder version

1.2.2

Fedora 36

-r

### Source repository URL

https://github.com/deltachat/deltachat-desktop.git

https://github.com/flathub/chat.delta.desktop.git

### Description

Observed when generating sources for deltachat desktop as of `v1.32.1`:
```
    assert (package.lockfile.parent / source.path / 'package.json').is_file()
AssertionError
```
This is cause by the following `@deltachat/jsonrpc-client` package lockfile entry:
```json
    "@deltachat/jsonrpc-client": {
      "version": "file:node_modules/deltachat-node/deltachat-jsonrpc/typescript",
      "requires": {
        ...
      }
    },
```
So, as I understand, it references a local source that isn't available before `npm install` is ran, thus flatpak-node-generator fails on a clean app source checkout.

gasinvein:
Fixes #312

### DIFF
--- a/node/flatpak_node_generator/providers/npm.py
+++ b/node/flatpak_node_generator/providers/npm.py
@@ -311,7 +311,7 @@ class NpmModuleProvider(ModuleProvider):
             self.gen.add_git_source(source.url, source.commit, path)
 
         elif isinstance(source, LocalSource):
-            assert (package.lockfile.parent / source.path / 'package.json').is_file()
+            pass
 
         else:
             raise NotImplementedError(


node/flatpak_node_generator/providers/npm.py
```diff
@@ -311,6 +311,8 @@ async def generate_package(self, package: Package) -> None:
             self.gen.add_git_source(source.url, source.commit, path)
 
         elif isinstance(source, LocalSource):
+            if Path('node_modules') in Path(source.path).parents:
```
refi64, line 314: Hmm, should we...maybe just remove the assert altogether? This feels like a slightly odd heuristic to me (ngl although I'm not sure how it ended up in the lockfile in the first place)

gasinvein: Well, ignoring local sources altogether doesn't feel right, but I guess we don't have anything to do with them anyway.

refi64:
whoops, didn't realize this already had pushed to it

PR #314, opened 2022-08-24
Allow using binary wheels instead of building from source

### LINKED ISSUE #296: Allow using binary wheels instead of building from source
### flatpak-builder version

1.0.10

Ubuntu 20.04

pip/flatpak-pip-generator

_No response_

### Source repository URL

_No response_

_No response_

### Description

When generating a dependency file for the python package `cryptography` from pypi.org, the `flatpak-pip-generator` recognised that the package only supplies platform dependent wheel distributions and therefore includes the source distribution . However the source distribution can only be build by compiling complicated rust dependencies and building the app with flatpak-builder fails therefore. I solved this by replacing the source distribution with all the different wheel packages available so that the right one is always available. Why can't `flatpak-pip-generator` do something similar?

PS: Which wheel distributions do I actually need for building the flatpak on aarch64 and x86?

real-yfprojects:
Resolves #296.

Some refactoring eases the process.

### DIFF
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -12,6 +12,7 @@ import sys
 import tempfile
 import urllib.request
 from collections import OrderedDict
+from typing import Any, Dict, List
 
 try:
     import requirements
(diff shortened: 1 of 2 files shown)


real-yfprojects:
This is closed in favour of [req2flatpak.py](https://github.com/johannesjh/req2flatpak). See https://github.com/flatpak/flatpak-builder-tools/issues/296#issuecomment-1279834172.

PR #315, opened 2022-08-26
node: Improve live progress output (and other small tweaks)

refi64:
Second commit message has the basic summary:

> This upgrades the progress output quite a bit:
> 
> - There's a pretty animation at the start of the line!
> - Instead of printing the last package interacted with (which made no
>  sense), show as many of the currently-processing packages as can fit
>  on a single line.
> - Show live download progress for large downloads, so any that are
>   holding up generation don't result in appears to be a hang.
> 
> In addition, this adds a hidden CLI flag, --traceback-on-interrupt, that
> prints a full traceback from every active package coroutine on Ctrl-C,
> which was useful while debugging this (and probably will be in the
> future!)
> 

Live demo (with a weirdly low framerate, *but it actually looks smooth live*, not sure why it was being weird here):

[Screencast from 08-26-2022 05:59:29 PM.webm](https://user-images.githubusercontent.com/1690697/187001473-3ef06323-9205-45bf-91f2-7ab647fb8468.webm)

This may or may not have taken like, 6 hours to figure out. (Terminals are hard.)

### DIFF
--- a/node/flatpak_node_generator/main.py
+++ b/node/flatpak_node_generator/main.py
@@ -6,6 +6,7 @@ import asyncio
 import json
 import os
 import sys
+import time
 
 from .cache import Cache, FilesystemBasedCache
 from .manifest import ManifestGenerator
@@ -187,6 +188,8 @@ async def _async_main() -> None:
     else:
         assert False, args.type
 
+    start_time = time.monotonic()
+
     print('Reading packages from lockfiles...')
     packages: Set[Package] = set()
     rcfile_node_headers: Set[NodeHeaders] = set()
@@ -246,6 +249,8 @@ async def _async_main() -> None:
             )
             gen.add_command(f'bash {gen.data_root / script_name}')
 
+    elapsed = round(time.monotonic() - start_time, 1)
+
     if args.split:
         i = 0
         for i, part in enumerate(gen.split_sources()):
@@ -254,7 +259,7 @@ async def _async_main() -> None:
             with open(output, 'w') as fp:
                 json.dump(part, fp, indent=ManifestGenerator.JSON_INDENT)
 
-        print(f'Wrote {gen.source_count} to {i + 1} file(s).')
+        print(f'Wrote {gen.source_count} to {i + 1} file(s) in {elapsed} second(s).')
     else:
         with open(args.output, 'w') as fp:
             json.dump(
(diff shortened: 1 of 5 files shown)


gasinvein:
There is a plenty or ready-made "progressbar" libs, honestly I'd prefer to use one of those instead of implementing (or extending) our own.

refi64:
Right, so I actually started out using tqdm but ended up removing it, because as it turns out, I didn't actually want a *progress* bar, per se, and it definitely wasn't a *single* line.

The main things are that, if you see in the video:

- The terminal space is used to show the packages being processed. I felt like this was a lot more useful, esp since it was clearer what packages might actually be holding things up. We can't fit as many on one line with a progress bar there
- Multiple progress lines might be spawned when packages need big files. Unfortunately, although tqdm handles nested progress bars well, it does *not* like it when a progress bar in the middle disappears before the ones around it:
  ```python
  from tqdm import tqdm
  import time
  
  p1 = tqdm(total=10)
  p2 = tqdm(total=10)
  p3 = tqdm(total=10)
  
  for bar in p2, p3, p1:
      for i in range(5):
          time.sleep(0.2)
          bar.update(i)
  
      bar.close()
  ```
  Here, after the middle bar closes, *four* lines appear on the screen, with one 100% line being repeated twice.
- There's no particular guarantee we'd have a length when downloading a file, and if we *don't*...tqdm's support for some display of infinite progress isn't that great.
- ...tqdm doesn't let me easily do the pretty animation.

The main loss is honestly time estimates, but given that some packages require extra downloads and such anyway, those weren't super useful.

Additionally, although there are a bunch of other libraries...they all seem to be missing something relatively tricky that we'd have to add anyway, most particularly the proper handling of multiple progress bars. Hence, DIY progress bars.

(Plus, the code isn't *that* untested, I ripped the core logic from an old tool in C++ that had multiple lines of progress that would finish at inconsistent intervals. Feel free to try to break it though!)

refi64:
So as it turns out, there *is* a nice library that does what I need, it's just not marketing as a progress library: [Rich](https://github.com/Textualize/rich). It also gives some *very* fancy and nice progress bars! So, the current output is:

[Screencast from 2022-12-16 16-44-59.webm](https://user-images.githubusercontent.com/1690697/208201722-cabf3dc9-04e1-42b9-86a0-f4a2ad27c2b3.webm)

(ignore the random frame drops at the 0:07 and 0:14 mark, no idea what those are from)

Ideally I'd like to add some colors which Rich makes relatively easy, *but* this is already a big improvement and that can be added later.

PR #317, opened 2022-09-18
node: Support custom registries

### LINKED ISSUE #316: flatpak-node-generator fails with sha1 integrity check
### flatpak-builder version

1.2.0

Ubuntu 20.04

--no-requests-cache --no-xdg-layout

### Source repository URL

https://raw.githubusercontent.com/cinnyapp/cinny/v2.2.0/package-lock.json

https://raw.githubusercontent.com/flathub/in.cinny.Cinny/master/in.cinny.Cinny.yml

### Description

I have updated my dependency (@matrix-org/olm) and it has sha1 integrity hash now. Now builder seems to be failing when processing that dependency to generate node sources.

Logs: https://github.com/flathub/in.cinny.Cinny/actions/runs/3073712103/jobs/4966018497#step:12:3399

gasinvein:
This is a rather quick-and-dirty implementation. We probably can implement custom registries in a nicer way, but at least this does the job.
Fixes #316

### DIFF
--- a/node/flatpak_node_generator/providers/npm.py
+++ b/node/flatpak_node_generator/providers/npm.py
@@ -13,6 +13,7 @@ from typing import (
 
 import asyncio
 import collections
+import functools
 import json
 import re
 import shlex
@@ -142,6 +143,8 @@ class NpmModuleProvider(ModuleProvider):
         self.git_sources: DefaultDict[
             Path, Dict[Path, GitSource]
         ] = collections.defaultdict(lambda: {})
+        # FIXME better pass the same provider object we created in main
+        self.rcfile_provider = NpmRCFileProvider()
 
     def __exit__(
         self,
@@ -209,7 +212,7 @@ class NpmModuleProvider(ModuleProvider):
             cache_future = asyncio.get_event_loop().create_future()
             self.registry_packages[package.name] = cache_future
 
-            data_url = f'{self.registry}/{package.name.replace("/", "%2f")}'
+            data_url = f'{self.get_package_registry(package)}/{package.name.replace("/", "%2f")}'
             # NOTE: Not cachable, because this is an API call.
             raw_data = await Requests.instance.read_all(data_url, cachable=False)
             data = json.loads(raw_data)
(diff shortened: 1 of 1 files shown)


node/flatpak_node_generator/providers/npm.py
```diff
@@ -321,6 +324,29 @@ async def generate_package(self, package: Package) -> None:
     def relative_lockfile_dir(self, lockfile: Path) -> Path:
         return lockfile.parent.relative_to(self.lockfile_root)
 
+    @functools.lru_cache(typed=True)
+    def get_lockfile_rc(self, lockfile: Path) -> Dict[str, str]:
+        rc = {}
+        rcfile_paths = []
+        for path in lockfile.parents:
```
refi64, line 331: Afaik npmrc files aren't evaluated up a tree, only yarnrc files?

node/flatpak_node_generator/providers/npm.py
```diff
@@ -142,6 +143,8 @@ def __init__(
         self.git_sources: DefaultDict[
             Path, Dict[Path, GitSource]
         ] = collections.defaultdict(lambda: {})
+        # FIXME better pass the same provider object we created in main
```
refi64, line 146: this hurts :cry: but it's tricky to change, I'm playing around a bit with it rn but we can definitely just clean it all up later.

node/flatpak_node_generator/providers/npm.py
```diff
@@ -321,6 +324,29 @@ async def generate_package(self, package: Package) -> None:
...
+                rcfile_paths.append(rcfile_path)
+        for rcfile_path in reversed(rcfile_paths):
+            rc.update(self.rcfile_provider.parse_rcfile(rcfile_path))
+        return rc
+
+    def get_package_registry(self, package: Package) -> str:
+        assert isinstance(package.source, RegistrySource)
+        rc = self.get_lockfile_rc(package.lockfile)
+        if rc and '/' in package.name:
+            org, _ = package.name.split('/', maxsplit=1)
```
refi64, line 345:
```suggestion
            scope, _ = package.name.split('/', maxsplit=1)
```

node/flatpak_node_generator/providers/npm.py
gasinvein: Not sure. Is it documented somewhere? Or any hints in npm source code?

node/flatpak_node_generator/providers/npm.py
gasinvein: It does. TBH while implementing this, I started to doubt if adding a separate factory class for rcfiles was a good idea at all. But ofc this is out of scope of this PR.

node/flatpak_node_generator/providers/npm.py
gasinvein: I'll just take your word on it. Even if it proves wrong, we can put back the tree lookup code.

PR #318, opened 2022-09-25
Add requirement about python3-devel

razzeee:
This was failing for me and I needed some time to find the problem, so I think it makes sense to mention it.

### DIFF
--- a/node/README.md
+++ b/node/README.md
@@ -12,6 +12,7 @@ required, however, you can disable it explicitly via `--no-xdg-layout`.
 
 - flatpak-builder 1.1.2 or newer
 - Python 3.7+.
+- Python dev package (install on fedora via `sudo dnf install python3-devel`)
 - [pipx](https://pypa.github.io/pipx/) (recommended) or
   [pip](https://pip.pypa.io/en/stable/) (both of these are usually available in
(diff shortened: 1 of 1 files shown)


refi64:
What was the exact problem you hit? All the generator's dependencies should be wheels, so I'm surprised you needed python3-devel...

razzeee:
If I remember right, it couldn't build a wheel as it was missing a header

razzeee:
I ended up on this SO https://stackoverflow.com/questions/34668675/cython-fatal-error-python-h-no-such-file-or-directory

refi64:
Hmm so I just tried this in a Fedora docker image, with no Python development files installed or even a C compiler, and it still installed...any chance you could try reinstalling it *without* the python3-devel package, and share the original error here?

razzeee:
Hrm, tried this in a fedora 36 vm I had laying around and it just worked. Without `python3-devel` installed, so I too am confused about this.

Should we close it?

refi64:
...out of curiosity, were you on the Fedora 37 beta? It would actually make sense there, because they shipped Python 3.11's rc versions, probably before wheels were readily available for some dependencies.

razzeee:
Yes, that's was/is the case.

PR #325, opened 2022-10-12
cargo: Allow filtering packages by name

aleb:
Cargo.lock might contain unwanted packages.

### DIFF
--- a/cargo/flatpak-cargo-generator.py
+++ b/cargo/flatpak-cargo-generator.py
@@ -287,7 +287,18 @@ async def get_package_sources(package, cargo_lock, git_repos):
     return (crate_sources, {'crates-io': {'replace-with': VENDORED_SOURCES}})
 
 
-async def generate_sources(cargo_lock, git_tarballs=False):
+def packages_dependencies(names, all_packages):
+    for name in names or ():
+        # Some are like "time 0.3.15"
+        name = name.split(' ')[0]
+        yield name
+        package = all_packages[name]
+        deps = package.get('dependencies')
+        for dep in packages_dependencies(deps, all_packages):
+            yield dep
+
+
+async def generate_sources(cargo_lock, packages=None, git_tarballs=False):
     # {
     #     "git-repo-url": {
     #         "lock": asyncio.Lock(),
@@ -305,12 +316,22 @@ async def generate_sources(cargo_lock, git_tarballs=False):
         VENDORED_SOURCES: {'directory': f'{CARGO_CRATES}'},
     }
 
-    pkg_coros = [get_package_sources(p, cargo_lock, git_repos) for p in cargo_lock['package']]
+    if packages:
+        packages = packages.split(',')
+        logging.debug('Considering only packages: %s', packages)
+        all_packages = {p['name']: p for p in cargo_lock['package']}
+        packages = list(packages_dependencies(packages, all_packages))
+    else:
+        logging.debug('Considering all packages')
+        packages = list([p['name'] for p in cargo_lock['package']])
+
+    pkg_coros = [get_package_sources(p, cargo_lock, git_repos)
+                 for p in cargo_lock['package']
+                 if p['name'] in packages]
     for pkg in await asyncio.gather(*pkg_coros):
         if pkg is None:
             continue
-        else:
-            pkg_sources, cargo_vendored_entry = pkg
+        pkg_sources, cargo_vendored_entry = pkg
         package_sources.extend(pkg_sources)
         cargo_vendored_sources.update(cargo_vendored_entry)
 
@@ -338,6 +359,7 @@ async def generate_sources(cargo_lock, git_tarballs=False):
 def main():
     parser = argparse.ArgumentParser()
     parser.add_argument('cargo_lock', help='Path to the Cargo.lock file')
+    parser.add_argument('-p', '--packages', help='Comma-separated packages in Cargo.lock to build')
     parser.add_argument('-o', '--output', required=False, help='Where to write generated sources')
     parser.add_argument('-t', '--git-tarballs', action='store_true', help='Download git repos as tarballs')
     parser.add_argument('-d', '--debug', action='store_true')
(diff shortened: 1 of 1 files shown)


gasinvein:
Can you please explain when and why it could be useful?
`Cargo.lock` contains complete list of the project's dependencies, and omitting any of them is expected to result in a broken build. So, why someone would want that is definitely not obvious.

aleb:
For example out of https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs we want only package "gst-plugin-gtk4", nothing else.

gasinvein:
I don't see a lockfile in the project you've linked, where does the `Cargo.lock` come from?

aleb:
https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html

> If you’re building a non-end product, such as a rust library that other rust [packages](https://doc.rust-lang.org/cargo/appendix/glossary.html#package) will depend on, put Cargo.lock in your .gitignore.

aleb:
You can `cargo fetch` to create the `Cargo.lock` file when it's missing.

gasinvein:
I still fail to understand what you're trying to achieve and what could be the reason for excluding cargo dependencies from a lockfile.

Just to be clear, to avoid possible misunderstandings: the `flatpak-cargo-generator.py` script doesn't build anything, neither cause anything to be installed. It only pre-fetches cargo packages cache (via flatpak-builder), very similar to what `cargo vendor` does (and largely replicating its behaviour).

aleb:
For example out of https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs we want only package "gst-plugin-gtk4", nothing else. 

```
gst-plugins-rs $ cargo build --package 2>&1 | grep gst | sort
    gst-plugin-audiofx
    gst-plugin-aws
    gst-plugin-cdg
    gst-plugin-claxon
    gst-plugin-closedcaption
    gst-plugin-csound
    gst-plugin-dav1d
    gst-plugin-fallbackswitch
    gst-plugin-ffv1
    gst-plugin-file
    gst-plugin-flavors
    gst-plugin-fmp4
    gst-plugin-gif
    gst-plugin-gtk4
    gst-plugin-hlssink3
    gst-plugin-hsv
    gst-plugin-json
    gst-plugin-lewton
    gst-plugin-onvif
    gst-plugin-raptorq
    gst-plugin-rav1e
    gst-plugin-regex
    gst-plugin-reqwest
    gst-plugin-rspng
    gst-plugin-rtpav1
    gst-plugin-sodium
    gst-plugin-spotify
    gst-plugin-textahead
    gst-plugin-textwrap
    gst-plugin-threadshare
    gst-plugin-togglerecord
    gst-plugin-tracers
    gst-plugin-tutorial
    gst-plugin-uriplaylistbin
    gst-plugin-version-helper
    gst-plugin-videofx
    gst-plugin-webp
    gst-plugin-webrtchttp
```

gasinvein:
From a workspace project like this, you select package(s) to build via cargo. Again, `flatpak-cargo-generator.py` doesn't control what is being built in any way. You can generate sources for the whole `gst-plugins-rs` project, build still build only desired components.

aleb:
Without the filtering we'd have to downloading a lot of stuff that's not needed, which is wasteful.

gasinvein:
`cargo vendor` doesn't consider "redundant" packages and downloads all the project dependencies, why should we do otherwise?

aleb:
Because we can! :)

`cargo-vendor` "Vendor all dependencies for a project locally". Why does it not allow to vendor only the deps of a single package? That I don't know.

As I said, gst-plugins-rs contains multiple libraries but we only need [to build] only one of them, not all. That's why we download only the dependencies for the corresponding package, not for all the packages in the project.

cargo/flatpak-cargo-generator.py
```diff
@@ -287,7 +287,18 @@ async def get_package_sources(package, cargo_lock, git_repos):
...
-async def generate_sources(cargo_lock, git_tarballs=False):
+def packages_dependencies(names, all_packages):
+    for name in names or ():
+        # Some are like "time 0.3.15"
+        name = name.split(' ')[0]
+        yield name
+        package = all_packages[name]
+        deps = package.get('dependencies')
+        for dep in packages_dependencies(deps, all_packages):
+            yield dep
```
gasinvein, line 298:
```suggestion
        yield from packages_dependencies(deps, all_packages)
```

cargo/flatpak-cargo-generator.py
```diff
@@ -305,12 +316,22 @@ async def generate_sources(cargo_lock, git_tarballs=False):
         VENDORED_SOURCES: {'directory': f'{CARGO_CRATES}'},
     }
 
-    pkg_coros = [get_package_sources(p, cargo_lock, git_repos) for p in cargo_lock['package']]
+    if packages:
+        packages = packages.split(',')
+        logging.debug('Considering only packages: %s', packages)
+        all_packages = {p['name']: p for p in cargo_lock['package']}
```
gasinvein, line 322: The `all_packages` dict feels out of place here, since it's only used within `packages_dependencies()` scope. Better pass the whole `cargo_lock` to `packages_dependencies()`, and there get whatever information is needed from it.

gasinvein (changes requested):

bbhtt:
> gst-plugins-rs contains multiple libraries

The upstream recommended way is to use the individual crates from crates.io https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/issues/1765

The `cargo.lock` inside the crates has dependencies for that only, afair.

```sh
wget https://static.crates.io/crates/gst-plugin-audiofx/gst-plugin-audiofx-0.13.4.crate
tar xf gst-plugin-audiofx-0.13.4.crate
cd gst-plugin-audiofx-0.13.4
python flatpak-cargo-generator.py Cargo.lock
```

bbhtt:
I guess this is not needed after all.

Issue #328, opened 2022-10-18
Unable to download `tensorboard_data_server-0.6.1`


TheEvilSkeleton:
### flatpak-builder version

1.2.2

### Linux distribution and version

Fedora 36 toolbx container (on Fedora 37 Silverblue host)

### Affected flatpak-builder tool

pip/flatpak-pip-generator

### Source repository URL

https://github.com/xinntao/Real-ESRGAN

### Description

Running `flatpak-pip-generator` against Real-ESRGAN's [`requirements.txt`](https://github.com/xinntao/Real-ESRGAN/blob/5ca1078535923d485892caee7d7804380bfc87fd/requirements.txt) fails to generate the manifest with the following error:
```
========================================================================
Downloading arch independent packages
========================================================================
Deleting MarkupSafe-2.1.1-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
Downloading https://files.pythonhosted.org/packages/1d/97/2288fe498044284f39ab8950703e88abbac2abbdf65524d576157af70556/MarkupSafe-2.1.1.tar.gz
Deleting kiwisolver-1.4.4-cp310-cp310-manylinux_2_12_x86_64.manylinux2010_x86_64.whl
Downloading https://files.pythonhosted.org/packages/5f/5c/272a7dd49a1914f35cd8d6d9f386defa8b047f6fbd06badd6b77b3ba24e7/kiwisolver-1.4.4.tar.gz
Deleting contourpy-1.0.5-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
Downloading https://files.pythonhosted.org/packages/38/b3/d6fd43ab2eadce72ac089328d80e9cdf274efdb79a9933aaf52ef1621e99/contourpy-1.0.5.tar.gz
Deleting matplotlib-3.6.1-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
Downloading https://files.pythonhosted.org/packages/0f/23/2eed6f40e1afca9c272fa7c13b437b7784c94a0c180f796c0da114265bf9/matplotlib-3.6.1.tar.gz
Traceback (most recent call last):
  File "/var/home/TheEvilSkeleton/flatpak-builder-tools/pip/flatpak-pip-generator", line 259, in <module>
    url = get_tar_package_url_pypi(name, version)
[...]

PR #331, opened 2022-10-31
Allow arch dependent packages

AndersBlomdell:
Adds two new options to flatpak-pip-generator:

  --arch-dependent-allow  Allow arch dependent packages to be used
                          if no arch-independent package can be found.
                          https://github.com/flatpak/flatpak-builder-tools/issues/328
  --arch-dependent-force  Force arch dependent packages to always be used,
                          Useful if the Sdk doesn't have the tools to rebuild
                          the arch independent package

### DIFF
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -44,6 +44,10 @@ parser.add_argument('--output', '-o',
                     help='Specify output file name')
 parser.add_argument('--runtime',
                     help='Specify a flatpak to run pip inside of a sandbox, ensures python version compatibility')
+parser.add_argument('--arch-dependent-allow', action='store_true',
+                    help='Allow arch dependent packages to be used')
+parser.add_argument('--arch-dependent-force', action='store_true',
+                    help='Force arch dependent packages to be used')
 parser.add_argument('--yaml', action='store_true',
                     help='Use YAML as output format instead of JSON')
 opts = parser.parse_args()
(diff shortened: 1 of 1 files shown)


pip/flatpak-pip-generator
```diff
@@ -44,6 +44,10 @@ parser.add_argument('--output', '-o',
                     help='Specify output file name')
 parser.add_argument('--runtime',
                     help='Specify a flatpak to run pip inside of a sandbox, ensures python version compatibility')
+parser.add_argument('--arch-dependent-allow', action='store_true',
+                    help='Allow arch dependent packages to be used')
+parser.add_argument('--arch-dependent-force', action='store_true',
+                    help='Force arch dependent packages to be used')
```
ENDrain, line 50: README.md needs an update to mention the new parameters

pip/flatpak-pip-generator
```diff
@@ -251,19 +255,29 @@ with tempfile.TemporaryDirectory(prefix=tempdir_prefix) as tempdir:
...
-            url = get_tar_package_url_pypi(name, version)
-            print('Deleting', filename)
-            try:
-                os.remove(os.path.join(tempdir, filename))
-            except FileNotFoundError:
-                pass
-            print('Downloading {}'.format(url))
-            download_tar_pypi(url, tempdir)
+    if opts.arch_dependent_force:
+        fprint('Not using arch independent packages [--arch-dependent-force]')
```
ENDrain, line 259: If I'm reading this right, no code gets executed downloading any packages after this, arch dependent or otherwise. Don't suppose this is expected behavior

pip/flatpak-pip-generator
```diff
@@ -251,19 +255,29 @@ with tempfile.TemporaryDirectory(prefix=tempdir_prefix) as tempdir:
...
+                version = get_file_version(filename)
+                name = get_package_name(filename)
+                try:
+                    url = get_tar_package_url_pypi(name, version)
+                except Exception as e:
+                    if not opts.arch_dependent_allow:
+                        raise
+                    else:
+                        print('Ignoring "{}" [--arch-dependent-allow]'.format(e))
+                        continue
```
ENDrain, line 273: Again if I'm reading this right, no code gets executed downloading arch dependent packages after that. The rest of the loop is just skipped with `continue`, including `download_tar_pypi(url, tempdir)`. Don't suppose this is expected behavior either

ENDrain (changes requested):

pip/flatpak-pip-generator
ENDrain, line 50: README.md should have entries on these new parameters

ENDrain: Question: shouldn't this have options to choose which arches are we targeting? Likes of `arches-only` and `skip-arches`?

Otherwise LGTM.

ENDrain (changes requested):

bbhtt:
Hi @TingPing, can you please review this? A few latest submissions on Flathub are hitting the issue while trying to use pip-generator for pyside6/shiboken

pip/flatpak-pip-generator
bbhtt: > Question: shouldn't this have options to choose which arches are we targeting? Likes of `arches-only` and `skip-arches`?
> 
> Otherwise LGTM.

That seems complicated and a future work. Currently, pip download cannot download a package for another arch unless you specify the whole pep 600 platform tag, eg. `pip download --platform=manylinux_2_31_aarch64 --only-binary=:all: shiboken6`

The tag varies from package to package due to the glibc version/package availability.

There's also an issue where pip download doesn't handle legacy tag aliases too well https://github.com/pypa/pip/issues/10760

I think for now it should be left to host/pip.

hfiguiere:
this needs to be rebased. And it would be nice to have more info about which cases make this relevant.

hfiguiere:
#328 having been fixed #435 I wonder if this is still relevant.

PR #332, opened 2022-12-05
pip: Add support for pyproject.toml

JakobDev:
This PR adds support for reading [dependencies from a ypproject.toml](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#declaring-project-metadata). On Python versions older than 3.11, you need to install tomli to use this feature.

### DIFF
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -199,7 +199,12 @@ elif opts.pyproject_file:
     pyproject_file = os.path.expanduser(opts.pyproject_file)
     with open(pyproject_file, "rb") as f:
         pyproject_data = toml_load(f)
-    dependencies = pyproject_data.get("project", {}).get("dependencies", [])
+    is_poetry = pyproject_data.get('tool', {}).get('poetry') is not None
+    if is_poetry:
+        dependencies = pyproject_data.get('tool', {}).get('poetry', {}).get('dependencies', {})
+        dependencies.pop('python', None)
+    else:
+        dependencies = pyproject_data.get("project", {}).get("dependencies", [])
     packages = list(requirements.parse('\n'.join(dependencies)))
     with tempfile.NamedTemporaryFile('w', delete=False, prefix='requirements.') as req_file:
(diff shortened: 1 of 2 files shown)


pip/flatpak-pip-generator
```diff
@@ -158,6 +169,16 @@ if opts.requirements_file:
     except FileNotFoundError:
         pass
 
+elif opts.pyproject_file:
```
TingPing, line 172: Maybe it should error if both `--requirements-file` and `--pyproject-file` are given. Also please rebase and update `README.md` to include this option.

pip/flatpak-pip-generator
```diff
@@ -28,6 +28,8 @@ parser.add_argument('--cleanup', choices=['scripts', 'all'],
                     help='Select what to clean up after build')
 parser.add_argument('--requirements-file', '-r',
                     help='Specify requirements.txt file')
+parser.add_argument('--pyproject-file',
```
TingPing, line 31: Please rebase and update `README.md` to include this option.

ENDrain:
Hello, I've added suggested improvements to the original PR, hope @JakobDev reacts to this soon!

JakobDev:
I forget about this PR. Sorry for that.

@ENDrain Thanks, I have now merged your PR

ENDrain:
@TingPing can you give this an update please?

tpikonen:
This looks like a useful addition. Could you untangle the version history by removing the merges and rebase to current master?

Does support recursive dependencies which use `pyproject.toml`? That is, if a project has dependencies listed in `requirements.txt` but one of the dependencies uses `pyproject.toml` only, do I get all the dependencies in the generated file?

hfiguiere:
Please provide example of Python modules that need this.

tpikonen:
I ran into this with yt-dlp, which removed `requirements.txt` a while back.

hfiguiere:
providing the exact command that fail helps.

bbhtt:
Hrm, just remembered poetry v2 now supports PEP 621

PR #336, opened 2022-12-15
node: Add initial lockfile v3 support

### LINKED ISSUE #333: Support for lockfileVersion 3 in flatpak-node-generator
### flatpak-builder version

1.2.3

Fedora 37

_No response_

### Source repository URL

_No response_

_No response_

### Description

It seems that lockfileVersion 3 is not supported.

```bash
[ziga@ziga Passky-Desktop]$ flatpak-node-generator npm package-lock.json -o generated-sources.json
Reading packages from lockfiles...
  File "/home/ziga/.local/bin/flatpak-node-generator", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/ziga/.local/lib/python3.11/site-packages/flatpak_node_generator/main.py", line 277, in main
    asyncio.run(_async_main())
  File "/usr/lib64/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/asyncio/base_events.py", line 650, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/home/ziga/.local/lib/python3.11/site-packages/flatpak_node_generator/main.py", line 198, in _async_main
    packages.update(lockfile_provider.process_lockfile(lockfile))
  File "/home/ziga/.local/lib/python3.11/site-packages/flatpak_node_generator/providers/npm.py", line 101, in process_lockfile
    assert data['lockfileVersion'] <= 2, data['lockfileVersion']
AssertionError: 3
```

gasinvein:
There are 2 kinds of NPM lockfile syntaxes with major differences: modern, stored in `packages` NPM lockfile property, and legacy, stored in `dependencies`. I'm calling them `v2` and `v1`, respectively. Lockfiles version 1 contains only `v1`, version 2 contains both, and the recent version 3 contains only `v2`.

The `v2` syntax seems concise and overall more sane than `v1`. It's non-recursive, `version` field contains just the version, and, to my understanding the package origin is determined based on the URI scheme in the `resolved` field. However, the `v2` syntax doesn't seem provide package name information (the `name` field is optional and not present for most dependencies).

I didn't yet figure out how to adapt the git-source-patching code for `v2` syntax, thus git sources and unsupported for now, and thus the `v2` codepath is used only for lockfile version 3. Once we sort this out, we can process `v2` in lockfiles version 2, too.

Fixes #333

### DIFF
--- a/.github/workflows/node.yaml
+++ b/.github/workflows/node.yaml
@@ -44,7 +44,7 @@ jobs:
         run: |
           flatpak --user remote-add flathub https://flathub.org/repo/flathub.flatpakrepo
           flatpak --user install -y flathub \
-            org.freedesktop.{Platform,Sdk{,.Extension.node{14,16}}}//21.08
+            org.freedesktop.{Platform,Sdk{,.Extension.node{14,16}}}//22.08
       - name: Install dependencies
         run: poetry install
       - name: Run checks
--- a/.github/workflows/node.yaml
+++ b/.github/workflows/node.yaml
@@ -44,7 +44,7 @@ jobs:
         run: |
           flatpak --user remote-add flathub https://flathub.org/repo/flathub.flatpakrepo
           flatpak --user install -y flathub \
-            org.freedesktop.{Platform,Sdk{,.Extension.node{14,16}}}//22.08
+            org.freedesktop.{Platform,Sdk{,.Extension.node{14,16,18}}}//22.08
       - name: Install dependencies
         run: poetry install
       - name: Run checks
--- a/node/flatpak_node_generator/providers/npm.py
+++ b/node/flatpak_node_generator/providers/npm.py
@@ -52,10 +52,10 @@ class NpmLockfileProvider(LockfileProvider):
     def __init__(self, options: Options):
         self.no_devel = options.no_devel
 
-    def process_dependencies(
-        self, lockfile: Path, dependencies: Dict[str, Dict[Any, Any]]
+    def _process_packages_v1(
+        self, lockfile: Path, entry: Dict[str, Dict[Any, Any]]
     ) -> Iterator[Package]:
-        for name, info in dependencies.items():
+        for name, info in entry.get('dependencies', {}).items():
             if info.get('dev') and self.no_devel:
                 continue
             elif info.get('bundled'):
@@ -92,15 +92,18 @@ class NpmLockfileProvider(LockfileProvider):
             yield Package(name=name, version=version, source=source, lockfile=lockfile)
 
             if 'dependencies' in info:
-                yield from self.process_dependencies(lockfile, info['dependencies'])
+                yield from self._process_packages_v1(lockfile, info)
 
     def process_lockfile(self, lockfile: Path) -> Iterator[Package]:
         with open(lockfile) as fp:
             data = json.load(fp)
 
-        assert data['lockfileVersion'] <= 2, data['lockfileVersion']
-
-        yield from self.process_dependencies(lockfile, data.get('dependencies', {}))
+        if data['lockfileVersion'] in {1, 2}:
+            yield from self._process_packages_v1(lockfile, data)
+        else:
+            raise NotImplementedError(
+                f'Unknown lockfile version {data["lockfileVersion"]}'
+            )
 
 
 class NpmRCFileProvider(RCFileProvider):
--- a/node/flatpak_node_generator/providers/npm.py
+++ b/node/flatpak_node_generator/providers/npm.py
@@ -94,12 +94,76 @@ class NpmLockfileProvider(LockfileProvider):
             if 'dependencies' in info:
                 yield from self._process_packages_v1(lockfile, info)
 
+    def _process_packages_v2(
+        self, lockfile: Path, entry: Dict[str, Dict[Any, Any]]
+    ) -> Iterator[Package]:
+        for install_path, info in entry.get('packages', {}).items():
+            if (info.get('dev') or info.get('devOptional')) and self.no_devel:
+                continue
+            if info.get('link'):
+                # NOTE We're not interested in symlinks, NPM will create them at install time
+                # but we still could collect package symlinks anyway just for completeness
+                continue
+
+            name = info.get('name')
+
+            source: PackageSource
+            package_json_path = lockfile.parent / install_path / 'package.json'
+            if (
+                'node_modules' not in package_json_path.parents
+                and package_json_path.exists()
+            ):
+                source = LocalSource(path=install_path)
+                if name is None:
+                    with package_json_path.open('rb') as fp:
+                        name = json.load(fp)['name']
+            elif 'resolved' in info:
+                resolved_url = urllib.parse.urlparse(info['resolved'])
+                if resolved_url.scheme == 'file':
+                    source = LocalSource(path=resolved_url.path)
+                elif resolved_url.scheme in {'http', 'https'}:
+                    integrity = Integrity.parse(info['integrity'])
+                    # NOTE I don't know how to determine if the source came from a registry
+                    # based on the lockfile alone, so unconditionally handling it as if it was
+                    # a "URL as dependency" kind of source
+                    source = PackageURLSource(
+                        integrity=integrity, resolved=info['resolved']
+                    )
+                elif resolved_url.scheme.startswith('git+'):
+                    raise NotImplementedError(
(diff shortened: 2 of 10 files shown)


.github/workflows/node.yaml
```diff
@@ -44,7 +44,7 @@ jobs:
         run: |
           flatpak --user remote-add flathub https://flathub.org/repo/flathub.flatpakrepo
           flatpak --user install -y flathub \
-            org.freedesktop.{Platform,Sdk{,.Extension.node{14,16}}}//21.08
+            org.freedesktop.{Platform,Sdk{,.Extension.node{14,16}}}//22.08
```
refi64: This should also update `tools/lockfile-utils.sh` (as well as adding node18 support).

node/tests/conftest.py
```diff
@@ -173,7 +174,12 @@ def package_json(self) -> Path:
     @property
     def lockfile_source(self) -> Path:
         if self.type == ProviderFactoryType.NPM:
-            suffix = self._V2_JSON if self.node_version >= 16 else self._V1_JSON
+            if self.node_version >= 18:
+                suffix = self._V3_JSON
+            elif self.node_version >= 16:
+                suffix = self._V2_JSON
+            else:
+                suffix = self._V1_JSON
```
refi64, line 182: super nit: newline after this would be nice since it's a more complex block now

.github/workflows/node.yaml
```diff
@@ -44,7 +44,7 @@ jobs:
         run: |
           flatpak --user remote-add flathub https://flathub.org/repo/flathub.flatpakrepo
           flatpak --user install -y flathub \
-            org.freedesktop.{Platform,Sdk{,.Extension.node{14,16}}}//22.08
+            org.freedesktop.{Platform,Sdk{,.Extension.node{14,16,18}}}//22.08
```
refi64, line 47: Does this actually include npm 9? I noticed the comments on the other issue + the release flow seem to imply it hadn't been backported yet...

node/flatpak_node_generator/providers/npm.py
```diff
@@ -94,12 +94,74 @@ def _process_packages_v1(
...
+                if not name:
+                    with package_json_path.open('rb') as fp:
+                        name = json.load(fp)['name']
+            elif 'resolved' in info:
+                resolved_url = urllib.parse.urlparse(info['resolved'])
+                if resolved_url.scheme == 'file':
+                    source = LocalSource(path=resolved_url.path)
+                elif resolved_url.scheme in {'http', 'https'}:
+                    integrity = Integrity.parse(info['integrity'])
+                    # NOTE I don't know how to determine if the source came from a registry
```
refi64, line 123: Wow, does npm not barf at the lack of cached metadata responses? If not that's actually pretty awesome, having to add that is really ugly.

node/flatpak_node_generator/providers/npm.py
```diff
@@ -94,12 +94,74 @@ def _process_packages_v1(
...
+                # but we still could collect package symlinks anyway just for completeness
+                continue
+
+            name = info.get('name')
+
+            source: PackageSource
+            package_json_path = lockfile.parent / install_path / 'package.json'
+            if package_json_path.exists():
+                source = LocalSource(path=install_path)
+                if not name:
```
refi64, line 114: nit: `if name is None` (ditto for the same test further down)

node/flatpak_node_generator/providers/npm.py
```diff
@@ -94,12 +94,74 @@ def _process_packages_v1(
...
+            if info.get('link'):
+                # NOTE We're not interested in symlinks, NPM will create them at install time
+                # but we still could collect package symlinks anyway just for completeness
+                continue
+
+            name = info.get('name')
+
+            source: PackageSource
+            package_json_path = lockfile.parent / install_path / 'package.json'
+            if package_json_path.exists():
```
refi64, line 112: Is there any chance this will trigger if someone has already run `npm install`, so a registry package would be in `node_modules`?

node/flatpak_node_generator/package.py
```diff
@@ -128,8 +128,10 @@ class LocalSource(PackageSource):
...
 
 
-class Package(NamedTuple):
+@dataclass(frozen=True, eq=True)
+class Package:
     name: str
     version: str
     source: PackageSource
     lockfile: Path
+    install_path: Optional[str] = field(default=None)
```
refi64, line 137: Do we actually need this field? I don't see it used anywhere. (If so, probably also just changing the default to `= None`, `field(...)` isn't really useful here.)

.github/workflows/node.yaml
gasinvein: No, node18 still has npm 8. I thought about bundling npm 9 at build time, but concluded that the fact that node18 understands lockfiles v3 is enough, since node is only reading lockfiles during our tests, and generated the lockfiles with npm 9 installed on host.

.github/workflows/node.yaml
gasinvein: Ugh, I've forgot about the script because I didn't use it. Done.

node/flatpak_node_generator/providers/npm.py
gasinvein: To my understanding, NPM doesn't need cached registry api responses as long as both `resolved` and `integrity` fields are present in the lockfile, but sadly it's not always the case.

node/flatpak_node_generator/providers/npm.py
gasinvein: It will. But running the generator against a dirty source tree will screw up the generated sources anyway, so I'm not sure if it makes sense to add a safety check just here.

node/flatpak_node_generator/package.py
gasinvein: No, not really. I've added it just for completeness.

jwillikers:
@gasinvein Thanks for this PR! It allowed me to update dependencies for flathub/net.hovancik.Stretchly#3 which now uses the V3 format.

node/flatpak_node_generator/providers/npm.py
refi64: oh that's unfortunate :sweat_smile: since when is that the case? I think it used to work at *some* point...

Not really an issue for this PR regardless, though.

node/flatpak_node_generator/providers/npm.py
refi64: Ah and that falls back to the "idk" case right now.

.github/workflows/node.yaml
refi64: Ah, it seems that [npm 9 is also going to make it to to node 18](https://github.blog/changelog/2022-10-24-npm-v9-0-0-released/#timeline-to-ga) makes this a non-issue anyway.

node/flatpak_node_generator/package.py
refi64: Personally I'd rather just remove it for now then, maybe make a comment note in `npm.py` that it's a field that exists that may be useful at some point? Otherwise it just looks like dead code.

gasinvein: Ok, removed it.

node/flatpak_node_generator/providers/npm.py
gasinvein: IIRC it always was the case. I mean, if a `package-lock.json` or `yarn.lock` file somehow gets into a release tarball in the registry, we'll be generating unused dependencies from there as well, potentially exploding the size of `generated-sources.json`.

gasinvein: In any case, added a condition check assuming that a local source can't be inside a `node_modules` directory.

gasinvein:
> I didn't yet figure out how to adapt the git-source-patching code for `v2` syntax

A note for the future: we just don't. NPM translates most of the git sources to hosted git service tarball urls, and pulls them like normal packages over http. Given that the vast majority of the git projects are hosted on several well-known git hosting providers, I'm convinced we should just re-implement the NPM "hosted git tarball" logic and avoid dealing with git altogether.

PR #338, opened 2022-12-25
cargo: fix crash if workspace is empty

gdesmott:
If for some reason workspace is empty, root_toml['workspace'] is {}, which was raising a KeyError.

This was triggered by librespot:
https://github.com/librespot-org/librespot/blob/dev/Cargo.toml#L13

### DIFF
--- a/cargo/flatpak-cargo-generator.py
+++ b/cargo/flatpak-cargo-generator.py
@@ -160,7 +160,7 @@ async def get_cargo_toml_packages(root_toml, root_dir):
         packages[root_toml['package']['name']] = root_dir
 
     if 'workspace' in root_toml:
-        for member in root_toml['workspace']['members']:
+        for member in root_toml['workspace'].get('members', []):
             for subpkg_toml in glob.glob(os.path.join(root_dir, member, 'Cargo.toml')):
                 subpkg = os.path.normpath(os.path.dirname(subpkg_toml))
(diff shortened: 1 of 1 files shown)


cargo/flatpak-cargo-generator.py
```diff
@@ -160,13 +160,15 @@ async def get_dep_packages(entry, toml_dir):
...
     if 'workspace' in root_toml:
-        for member in root_toml['workspace']['members']:
-            for subpkg_toml in glob.glob(os.path.join(root_dir, member, 'Cargo.toml')):
-                subpkg = os.path.normpath(os.path.dirname(subpkg_toml))
-                logging.debug("Loading workspace member %s in %s", member, root_dir)
-                pkg_toml = load_toml(subpkg_toml)
-                await get_dep_packages(pkg_toml, subpkg)
-                packages[pkg_toml['package']['name']] = subpkg
+        if 'members' in root_toml['workspace']:
+            for member in root_toml['workspace']['members']:
```
gasinvein, line 164: Please don't add another level of indentation, this function is too complex already. Maybe get a default empty list of 'members' instead?
```suggestion
        for member in root_toml['workspace'].get('members', []):
```

gdesmott: That works, here you go.

gasinvein:
Thanks!

PR #342, opened 2023-01-13
node: upgrade app templates

proletarius101:
including versions of
- SDK
- Node.js
- Electron

It's impossible to upgrade SDKs for `build.electron.webpack.ElectronWebpackQuickStart` without replacing the `package.json` due to `error:0308010C:digital envelope routines::unsupported`. So I keep it as is.


refi64:
> It's impossible to upgrade SDKs for build.electron.webpack.ElectronWebpackQuickStart without replacing the package.json due to error:0308010C:digital envelope routines::unsupported. 

Isn't it possible to work around this via some env variable? Or does that not work anymore...

PR #343, opened 2023-01-13
go: Add docs for module-aware build mode.

dennwc:
Current build method described in `go-get` section relies on legacy `GOPATH` mode.

Also, there's a `flatpak-go-vendor-generator.py` that generates `vendor` directory, but refers to https://github.com/golang/go/issues/35922 for a better way to extract direct URLs.

This change adds documentation for a fully modules-aware method (with examples) which supports all edge cases (like `replace` directives in `go.mod`) and downloads ZIP archives of the dependencies from the Go modules proxy.

It requires Go tooling on the host (specifically `go mod download -json`) and thus assumes that installing an external Go tools for generation is acceptable.


go-modules/README.md
```diff
@@ -0,0 +1,93 @@
...
+This method generates sources section with references to ZIP files hosted by the Go modules proxy.
+
+## Usage
+
+Example manifest (yaml):
+
+```yaml
+app-id: io.github.golang.tools.gorename
+runtime: org.freedesktop.Platform
+runtime-version: '22.08'
```
hfiguiere, line 13:
```suggestion
runtime-version: '23.08'
```

go-modules/README.md
```diff
@@ -0,0 +1,93 @@
...
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.golang
+command: echo "Done"
```
hfiguiere, line 17: this is not valid. it expect the name of an executable.

go-modules/README.md
```diff
@@ -0,0 +1,93 @@
...
+3. Append generated `go.mod.yaml` file to `sources` section of the manifest.
+4. Add `modules.txt` to your Flatpak project.
+5. Build/install your binary in vendor mode (with `-mod=vendor`).
+
+## Final manifest example
+
+```yaml
+app-id: io.github.golang.tools.gorename
+runtime: org.freedesktop.Platform
+runtime-version: '22.08'
```
hfiguiere, line 42:
```suggestion
runtime-version: '23.08'
```

go-modules/README.md
```diff
@@ -0,0 +1,93 @@
...
+sdk: org.freedesktop.Sdk
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.golang
+command: echo "Done"
```
hfiguiere, line 46: same as above.

hfiguiere:
```suggestion
command: gorename
```

I assume it's the name and it is installed.

go-modules/README.md
hfiguiere:
```suggestion
command: gorename
```

dennwc:
@hfiguiere Done! Rebased and applied the changes you mentioned.

PR #346, opened 2023-01-21
dotnet: Add flags for dotnet version and freedesktop SDK branch.

ProjectSynchro:
Adds the --freedesktop, -f, --dotnet, -d arguments to the script. Allows you to specify the dotnet version and which freedesktop sdk branch to use. Does not modify how the script behaves by default.

Makes the script more portable as you can specify what you're using and ensures that only valid versions can be selected.


ProjectSynchro:
Also updated the README slightly, outlining usable arguments without having to surf through the code.

dotnet/README.md
```diff
@@ -4,13 +4,20 @@ Tool to automatically generate a `flatpak-builder` sources file from a .NET Core
...
-you can use the sources file like this:
+Run `flatpak-dotnet-generator.py my-output-sources.json my.input.Desktop.csproj`. 
+
+### Arguments
+- `--runtime` or `-r` The target [runtime](https://learn.microsoft.com/en-us/dotnet/core/rid-catalog#linux-rids) to restore packages for.
+- `--dotnet` or `-d` The target version of dotnet to use. (5,6,7 are valid.)
+- `--freedesktop` or `-f` The target version of the freedesktop sdk to use. (21.08 by default, you can select 22.08 if wanted.)
+- `--destdir` The directory the generated sources file will save sources to `nuget-sources` by default. 
+
+Then, you can use the sources file like this:
```
refi64, line 20: This part should probably be under Usage still, the placement here feels odd.

dotnet/flatpak-dotnet-generator.py
```diff
@@ -17,6 +17,10 @@ def main():
     parser.add_argument('output', help='The output JSON sources file')
     parser.add_argument('project', help='The project file')
     parser.add_argument('--runtime', '-r', help='The target runtime to restore packages for')
+    parser.add_argument('--freedesktop', '-f', help='The target version of the freedesktop sdk to use', 
+                        choices=['21.08', '22.08'], default='21.08')
```
refi64, line 21: Any reason this defaults to `21.08` vs `22.08`? (Might be better to drop the choices too, so this doesn't need updates every runtime release.)

dotnet/flatpak-dotnet-generator.py
```diff
@@ -33,9 +37,9 @@ def main():
             'flatpak', 'run',
             '--env=DOTNET_CLI_TELEMETRY_OPTOUT=true',
             '--env=DOTNET_SKIP_FIRST_TIME_EXPERIENCE=true',
-            '--command=sh', '--runtime=org.freedesktop.Sdk//21.08', '--share=network',
-            '--filesystem=host', 'org.freedesktop.Sdk.Extension.dotnet6//21.08', '-c',
-            'PATH="${PATH}:/usr/lib/sdk/dotnet6/bin" LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/lib/sdk/dotnet6/lib" exec dotnet restore "$@"',
+            '--command=sh', '--runtime=org.freedesktop.Sdk//{}'.format(args.freedesktop), '--share=network',
+            '--filesystem=host', 'org.freedesktop.Sdk.Extension.dotnet{}//{}'.format(args.dotnet, args.freedesktop), '-c',
+            'PATH="${PATH}:'+'/usr/lib/sdk/dotnet{0}/bin" LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/lib/sdk/dotnet{0}/lib" exec dotnet restore "$@"'.format(args.dotnet),
```
refi64, line 42: Should be able to use f-strings for all of these (also using `${{PATH}}` to escape the brackets is cleaner than putting it in a separate string imo)

refi64 (changes requested):

dotnet/flatpak-dotnet-generator.py
ProjectSynchro: I chose to keep it at 21.08 to mirror how the script works atm, but may as well bump it to 22.08. Agreed on dropping the choices, at the time it made sense to make it near impossible to get a failure due to an incorrect runtime version.

dotnet/flatpak-dotnet-generator.py
ProjectSynchro: Good point. Haven't really written python since py2.6.. Goes to show the times have changed.

dotnet/README.md
ProjectSynchro: Agreed, at second glance this did look off.

dotnet/flatpak-dotnet-generator.py
ProjectSynchro: fixed as of 942b8e7.

dotnet/flatpak-dotnet-generator.py
ProjectSynchro: fixed as of d68dceb0c3686bed8fc75f59ed48ae73ad6dfc2d.

dotnet/README.md
ProjectSynchro: fixed as of 76b360d22d478738e4c2d450ec883a710c3bea37.

ProjectSynchro:
Apologies for not getting to this sooner, this didn't make it into my email.

let me know if there's any other changes that I should make, I think it should be good to go.

refi64:
Thanks!

PR #347, opened 2023-02-10
pip: Add --ignore-installed option

picsel2:
This option takes a comma-separated list of package names. It adds the `--ignore-installed` option to the pip command for these packages.

This is necessary for packages that are installed in the Sdk but then not part of the runtime (e.g. [lxml][1]).

[1]: https://gitlab.gnome.org/GNOME/gnome-build-meta/-/issues/380

### DIFF
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -38,6 +38,13 @@ parser.add_argument('--build-isolation', action='store_true',
                         'Mostly useful on pip that does\'t '
                         'support the feature.'
                     ))
+parser.add_argument('--ignore-installed',
+                    type=lambda s: s.split(','),
+                    default='',
+                    help='Comma-separated list of package names for which pip '
+                    'should ignore already installed packages. Useful when '
+                    'the package is installed in the SDK but not in the '
+                    'runtime.')
 parser.add_argument('--checker-data', action='store_true',
                     help='Include x-checker-data in output for the "Flatpak External Data Checker"')
 parser.add_argument('--output', '-o',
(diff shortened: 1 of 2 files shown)


pip/flatpak-pip-generator
```diff
@@ -38,6 +38,13 @@ parser.add_argument('--build-isolation', action='store_true',
                         'Mostly useful on pip that does\'t '
                         'support the feature.'
                     ))
+parser.add_argument('--ignore-installed',
+                    type=lambda s: s.split(","),
+                    default="",
+                    help="Comma-separated list of package names for which pip "
+                    "should ignore already installed packages. Useful when "
+                    "the package is installed in the SDK but not in the "
+                    "runtime.")
```
TingPing, line 47: Minor nit that this file mostly uses single quotes `'`.

Otherwise looks good to me.

picsel2: Good catch! I've changed it to single-quotes now.

dreua:
Note, that this only helps if lxml is installed as its own module. As soon as you are installing lxml as one of multiple sources for a module, i.e. sharing one pip install command, the situation ranges from complicates to impossible. I'm having this issue with pikepdf, see the [generated recipe](https://github.com/dreua/flatpak-module-qpdf-pikepdf/blob/940b8db012c3c7ed233c762fcf801bc92955402b/python3-pikepdf-internal.yaml) for an example.

1. To have any effect at all, you nee to set `--ignore-installed pikepdf` in this example, as there is no separate install command for lxml where it would be applied.
2. With this setting, however, we now have a problem with `packaging` which pikepdf also needs for building. Unlike `lxml`, `packaging` is present both in the Gnome SDK and runtime and it is filtered by a list hard coded in the pip-generator. This is all fine by itself, but combined with the `--ignore-installed` (or the `--force-reinstall` which I also tried manually) you land in a catch-22: The subsequent flatpak build fails with those flags, because pip does not consider any system packages then and fails because it can't find the `packaging` dependency.

Currently I simply ship with a dedicated `lxml` module duplicating the source. That works but it is annoying that there is no real solution for such issues, therefore I'm sharing this in the hope of finding a better one or at least keeping anyone else having this problem from going through the same unfruitful journey.

PR #353, opened 2023-03-11
poetry: support for poetry metadata format 2.0

GabMus:


ENDrain (approved):
LGTM. Maybe could also add a line in README.md listing metadata formats supported, but not sure if that's necessary. @mig5, since this worked for you could you also please add an approving review?

GabMus:
I really think we should merge this

d4nj1:
Would also be nice if the `--yaml` output format could be added similar to

https://github.com/flatpak/flatpak-builder-tools/blob/master/pip/flatpak-pip-generator#L55-L67
with
https://github.com/flatpak/flatpak-builder-tools/blob/master/pip/flatpak-pip-generator#L480-L491

hfiguiere:
Can you rebase this? We could get it merged eventually.

uninstall-your-browser:
`--production` does not seem to work, it still tries (and fails) to retrieve dev dependencies (unless I'm somehow generating my lock file wrong?). It should probably be changed to `--without <group-name>`, since poetry has deprecated `--no-dev`.

hfiguiere:
@GabMus unless this gets rebased, this can't be merged.

GabMus:
rebased, note that this PR is ancient and I don't even use flatpak-builder-tools anymore, so make sure this works as expected before merging

poetry/flatpak-poetry-generator.py
```diff
@@ -63,19 +63,23 @@ def get_module_sources(parsed_lockfile: dict, include_devel: bool = True) -> lis
...
+                    (
+                        package.get("category") == "dev"
+                        and include_devel
+                        and not package.get("optional")
+                    )
+                    or (
+                        package.get("category") == "main"
+                        and not package.get("optional")
+                    )
                 ):
```
bbhtt, line 76: the original logic and the rewritten one can be simplified a bit I think. It's a bit hard to follow this many ands/ors. Something like

```
category = package.get("category")
optional = package.get("optional", False)

if (
    not category
    or (category == "dev" and include_devel and not optional)
    or (category == "main" and not optional)
):
```

bbhtt:
> Would also be nice if the --yaml output format could be added similar to

This should be a separate issue. You just include the generated manifest. Having multiple formats is not a top priority.

> `--production` does not seem to work, it still tries (and fails) to retrieve dev dependencies

It should work now for v2 metadata.

>  Though it should probably be changed to `--without <group-name>`, since poetry has deprecated `--no-dev`.

Indeed, generic group support can be done later. This can be a separate issue. For now `--production` excludes things that have only `dev` in their groups.

poetry/flatpak-poetry-generator.py
bbhtt: Applied

bbhtt:
Tested this on a v1 and v2 lock files with the `--production` combination. Seems to work.

[poetryv2.txt](https://github.com/user-attachments/files/20025837/poetryv2.txt)
[poetryv1.txt](https://github.com/user-attachments/files/20025838/poetryv1.txt)

PR #355, opened 2023-04-09
node: skip errors on optional dependencies

proletarius101:
e.g. In https://raw.githubusercontent.com/lensapp/lens/master/package-lock.json

```json
    "node_modules/npm/node_modules/@colors/colors": {
      "version": "1.5.0",
      "inBundle": true,
      "license": "MIT",
      "optional": true,
      "engines": {
        "node": ">=0.1.90"
      }
    },
```

### DIFF
--- a/node/flatpak_node_generator/providers/npm.py
+++ b/node/flatpak_node_generator/providers/npm.py
@@ -134,6 +134,9 @@ class NpmLockfileProvider(LockfileProvider):
                         'Git sources in lockfile v2 format are not supported yet'
                         f' (package {install_path} in {lockfile})'
                     )
+            elif info.get('optional'):
+                # Optional dependencies can be skipped
+                continue
             else:
                 raise NotImplementedError(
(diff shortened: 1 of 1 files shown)


node/flatpak_node_generator/providers/npm.py
```diff
@@ -134,6 +134,9 @@ def _process_packages_v2(
                         'Git sources in lockfile v2 format are not supported yet'
                         f' (package {install_path} in {lockfile})'
                     )
+            elif info.get('optional'):
+                # Optional dependencies can be skipped
```
refi64, line 138: Should this at least be printing out a warning? Not sure if we do that anywhere else in the source yet, but it could be useful if the user was actually expecting the optional dep to still make its way inside.

bbhtt: @proletarius101 can you add a warning here saying something like "this {package} was skipped due to being optional"

Issue #265, opened 2022-04-08
pip: python generator produces empty sources


muelli:
I run the python generator against this requirements file: https://github.com/gnome-keysign/gnome-keysign/blob/1c5b04ac0b265491ab65b651e3bbbe0f4a310a18/requirements.txt ) and it produces empty sources:

```
$ flatpak-pip-generator --requirements-file=requirements.txt
========================================================================
Downloading sources
========================================================================
Running: "pip3 download --exists-action=i --dest /tmp/pip-generator-python3-requirements7hbolm42 -r /home/muelli/vcs/gnome-keysign/requirements.txt --require-hashes"
/usr/lib/python3/dist-packages/secretstorage/dhcrypto.py:15: CryptographyDeprecationWarning: int_from_bytes is deprecated, use int.from_bytes instead
  from cryptography.utils import int_from_bytes
/usr/lib/python3/dist-packages/secretstorage/util.py:19: CryptographyDeprecationWarning: int_from_bytes is deprecated, use int.from_bytes instead
  from cryptography.utils import int_from_bytes
Collecting attrs==21.4.0
  Using cached attrs-21.4.0-py2.py3-none-any.whl (60 kB)
Collecting autobahn[twisted]==22.3.2
  Using cached autobahn-22.3.2.tar.gz (376 kB)

...
...

Collecting zope.interface==5.4.0
  Using cached zope.interface-5.4.0-cp38-cp38-manylinux2010_x86_64.whl (259 kB)
Collecting setuptools
ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
    setuptools from https://files.pythonhosted.org/packages/0c/3c/548d361162702df85a0301f0cd0c47d05176b20bb086077a0fda740daf41/setuptools-62.0.0-py3-none-any.whl#sha256=a65e3802053e99fc64c6b3b29c11132943d5b8c8facbcc461157511546510967 (from autobahn[twisted]==22.3.2->-r requirements.txt (line 11))
Failed to download
Please fix the module manually in the generated file
========================================================================
Downloading arch independent packages
========================================================================
========================================================================
Obtaining hashes and urls
========================================================================
========================================================================
Generating dependencies
========================================================================
Generating dependencies for attrs

$ head python3-requirements.json
{
    "name": "python3-requirements",
    "buildsystem": "simple",
    "build-commands": [],
    "modules": [
        {
            "name": "python3-attrs",
            "buildsystem": "simple",
[...]

PR #357, opened 2023-04-12
pip: propagate error when downloading packages

muelli:
If pip fails to download, the generated sources file is empty and, arguably, useless. Instead of silently generating an empty file, the error is propagated now.

The root cause appears to be a bug in pip.

https://github.com/flatpak/flatpak-builder-tools/issues/265


hfiguiere:
can we get this merged? I keep getting hit by #265

Issue #362, opened 2023-08-12
name should be only the filename, not whole path


okias:
### flatpak-builder version

1.2.3

### Linux distribution and version

Debian 13 (trixie)

### Affected flatpak-builder tool

pip/flatpak-pip-generator

### Description

```
  python3 ./flatpak-pip-generator -r ../../python/requirements.txt \
    -o ../../support/flatpak/python3-requirements
```

leads to:
```
{
    "name": "../../support/flatpak/python3-requirements",
...
```

PR #363, opened 2023-08-12
pip: differenciate between package, path and filename properly

okias:
Fixes: https://github.com/flatpak/flatpak-builder-tools/issues/362

one additional changes is it uses only filename for the tmpfiles, since when directory contains `../` it makes a mess.


okias:
Thank you!

PR #364, opened 2023-08-13
Allow ignoring package in flatpak-pip-generator

bbhtt:
See commit message for explanation.

This is useful when someone wants to ignore a problematic package specified in `requirements.txt`. It can be due to multiple reasons that comes to mind: 

- the package is not available for current python version as a result downloading errors out even if other packages in the list are fine eg.

```
dataclasses==0.8
Sample2
```

- The package doesn't have a sdist tarball on pypi, and we want to manually add it to the manifest to build from source. Eg.

```
pyside6
Sample2
```

- Bugs like https://github.com/flatpak/flatpak-builder-tools/issues/328, that prevents producing a manifest altogether

(alternate take on https://github.com/flatpak/flatpak-builder-tools/pull/217, but a bit complicated)

Usage:

```
./flatpak-pip-generator --ignore-pkg <pkg1> <pkg2> -r reqs.txt
```

### DIFF
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -55,6 +55,8 @@ parser.add_argument('--yaml', action='store_true',
                     help='Use YAML as output format instead of JSON')
 parser.add_argument('--ignore-errors', action='store_true',
                     help='Ignore errors when downloading packages')
+parser.add_argument('--ignore-pkg', nargs='*',
+                    help='Ignore a package when generating the manifest. Can only be used with a requirements file')
 opts = parser.parse_args()
 
 if opts.yaml:
@@ -158,12 +160,20 @@ def fprint(string: str) -> None:
 
 packages = []
 if opts.requirements_file:
-    requirements_file = os.path.expanduser(opts.requirements_file)
+    requirements_file_input = os.path.expanduser(opts.requirements_file)
     try:
-        with open(requirements_file, 'r') as req_file:
+        with open(requirements_file_input, 'r') as req_file:
             reqs = parse_continuation_lines(req_file)
             reqs_as_str = '\n'.join([r.split('--hash')[0] for r in reqs])
-            packages = list(requirements.parse(reqs_as_str))
+            reqs_list = reqs_as_str.splitlines()
+            if opts.ignore_pkg:
+                reqs_new = '\n'.join(i for i in reqs_list if i not in opts.ignore_pkg)
+            else:
+                reqs_new = reqs_as_str
+            packages = list(requirements.parse(reqs_new))
+            with tempfile.NamedTemporaryFile('w', delete=False, prefix='requirements.') as req_file:
+                req_file.write(reqs_new)
+                requirements_file_output = req_file.name
     except FileNotFoundError:
         pass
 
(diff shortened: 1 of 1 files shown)


bbhtt:
One case that came up in #flatpak was requirements file with something like

```
dataclasses == 0.8; python_version < "3.7"
```

Quoting the whole string should work:

`--ignore-pkg 'dataclasses == 0.8; python_version < "3.7"' -r reqs.txt`

bbhtt:
> One case that came up in #flatpak was requirements file with something like
> 
> ```
> dataclasses == 0.8; python_version < "3.7"
> ```
> 
> Quoting the whole string should work:
> 
> `--ignore-pkg 'dataclasses == 0.8; python_version < "3.7"' -r reqs.txt`

Should work now without needing to quote the entire string.

pip/flatpak-pip-generator
```diff
@@ -158,22 +161,36 @@ def fprint(string: str) -> None:
...
+            reqs_list_raw = reqs_as_str.splitlines()
+            py_version_regex = re.compile(r';.*python_version .+$') # Remove when pip-generator can handle python_version
+            reqs_list = [py_version_regex.sub('', p) for p in reqs_list_raw]
+            if opts.ignore_pkg:
+                reqs_new_list = [i for i in reqs_list if i not in opts.ignore_pkg]
+                reqs_new = '\n'.join(reqs_new_list)
+            else:
+                reqs_new = reqs_as_str
+            packages = list(requirements.parse(reqs_new))
+            with tempfile.NamedTemporaryFile('w', delete=False, prefix='requirements.') as req_file:
```
TingPing, line 178: This just leaves temp files around. Move the scope up and let python delete it when the script exits.

pip/flatpak-pip-generator
```diff
@@ -158,22 +161,36 @@ def fprint(string: str) -> None:
...
+        with open(requirements_file_input, 'r') as req_file:
             reqs = parse_continuation_lines(req_file)
             reqs_as_str = '\n'.join([r.split('--hash')[0] for r in reqs])
-            packages = list(requirements.parse(reqs_as_str))
+            reqs_list_raw = reqs_as_str.splitlines()
+            py_version_regex = re.compile(r';.*python_version .+$') # Remove when pip-generator can handle python_version
+            reqs_list = [py_version_regex.sub('', p) for p in reqs_list_raw]
+            if opts.ignore_pkg:
+                reqs_new_list = [i for i in reqs_list if i not in opts.ignore_pkg]
+                reqs_new = '\n'.join(reqs_new_list)
```
TingPing, line 174: Minor style nit:

```suggestion
                reqs_new = '\n'.join(i for i in reqs_list if i not in opts.ignore_pkg)
```

pip/flatpak-pip-generator
bbhtt: Global variable or moving it up - neither works. I'm not exactly sure why.

I can add it as a global variable so that it doesn't need repeating for the `elif opts.packages` part, if you want.

What works is `os.remove` -ing it after `subprocess.run` is done. Right now, it is only called if an error occurs. `

pip/flatpak-pip-generator
bbhtt: Thanks!

pip/flatpak-pip-generator
TingPing: Oh I missed that this was already done elsewhere.

It is fine for now I suppose.

bbhtt: Handled now

PR #369, opened 2023-09-09
Go: Completely  rewrite

micahflee:
The existing Flatpak Go Get Generator no longer works with the modern Go ecosystem. The [flatpak-go-get-generator.py](https://github.com/flatpak/flatpak-builder-tools/blob/master/go-get/flatpak-go-get-generator.py) script works by looking for git repos in the Flatpak build dir after running `go get`, but that doesn't even create git repos anymore.

So this PR deletes the old `go-get` folder and makes a new `go` folder. This requires that Go and git are installed on the local system.

This script works by creating a new Go module in a temporary folder, add the given Go package as a dependency, and then runs `go list -m all` to get the full list of Go modules. For each module, it uses `go list -m -json <module>` to get detailed information. And then finally, it outputs the module in YAML format.


go/README.md
```diff
@@ -0,0 +1,102 @@
...
+build-options:
+  env:
+    GOBIN: /app/bin/
+buildsystem: simple
+name: meek-client
+sources:
+- dest: src/git/torproject/org/pluggable-transports/goptlib/git
+  tag: v1.1.0
+  type: git
+  url: https://git.torproject.org/pluggable-transports/goptlib.git.git
```
hfiguiere, line 65: shall the commit also be added to protect against retagging?

go/flatpak-go-deps.py
```diff
@@ -0,0 +1,282 @@
...
+            )
+            os.chdir(f"src/{repo_name}")
+
+            if version:
+                print(f"✨ Checking out version {version}")
+                subprocess.run(["git", "checkout", version], check=True)
+
+            os.chdir(temp_dir)
+
+            if folder:
```
jgillich, line 206: I think this `if` is a mistake. When given a root dir such as `github.com/a/b.git/`, this script detects no dependencies because it never switches into the repo directory. When I comment this line, it works.

Also it'd be nice if we could just give it the git url without the slash

PR #374, opened 2023-09-28
pip: adapt generator to be imported by other files

jacobszpz:
Just throwing an idea out here, but I have a use case where functions such as 'get_pypi_url' are useful to me and I am already using this script anyway. However, it is impossible to import from other files because there is no if __name__ == '__main__' clause in the script, as well as the script having no .py extension and having dashes in its name. I don't know if this aligns with your goals, but maybe you'd like to consider it.


jacobszpz:
For compatibility's sake, you could keep the original script as a wrapper and import the new version using exec

jacobszpz:
I also have added an option to the CLI to only download the main packages listed and not fetch all their dependencies

TingPing:
You will need to update all documentation.

I think its a slight downgrade for normal users but... its pretty minor.

jacobszpz:
Nah, I agree. I ended up not using this particular method anyway.

PR #375, opened 2023-10-06
Update node-gyp installVersion

inukshuk:
Latest version is 11, see https://github.com/nodejs/node-gyp/blob/main/package.json#L15

A lower version makes node-gyp remove the headers and force a new download

After a quick glance I think the changes for versions 10 and 11 affect the windows platform and error detection during tarball extraction so I think there's no other change required here.


refi64:
Could this cause problems for existing projects that worked w/ version `9`, or not really?

inukshuk:
I doubt it, because the changes seem to relevant to the Windows platform only, so they shouldn't affect flatpak builds on Linux at all.

PR #376, opened 2023-10-07
node: Create integrity file symlink for new electron/get versions

### LINKED ISSUE #370: @electron/get > v2.0.3 requires SHA256SUMS.txt for cached files
### flatpak-builder version

1.3.3

Fedora Linux 38.20230903.0 (Silverblue)

_No response_

### Source repository URL

_No response_

_No response_

### Description

`@electron/get` introduced a change to check the checksum file for cached files (https://github.com/electron/get/pull/212).
It looks for the checksum file at  `$XDG_CACHE_HOME/electron/$URL_HASH/SHA256SUMS.txt`

I think this is where the symlinks can be added

https://github.com/flatpak/flatpak-builder-tools/blob/de56f4702638739f930f4afa648686f12ac4d724/node/flatpak_node_generator/providers/special.py#L79-L85

I manually added the symlinks for https://github.com/flathub/com.logseq.Logseq/pull/93 and the build succeeded.

refi64:
This also updates the existing electron test & extends it to test this out.

Fixes #370.

### DIFF
--- a/node/flatpak_node_generator/electron.py
+++ b/node/flatpak_node_generator/electron.py
@@ -1,5 +1,9 @@
 from typing import Dict, Iterator, NamedTuple, Optional
 
+import hashlib
+import os.path
+import urllib.parse
+
 from .integrity import Integrity
 from .package import SemVer
 from .requests import Requests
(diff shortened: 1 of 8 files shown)


hfiguiere:
Maybe we could merge these....

L11R:
So could we merge that? Currently it's not usable with Electron apps without this patch.

Issue #377, opened 2023-10-13
[node] npm provider's `generated-sources.json` missing many packages (ENOTCACHED)


Ian2020:
### flatpak-builder version

1.3.3

### Linux distribution and version

Fedora 38

### Affected flatpak-builder tool

node/flatpak-node-generator.py

### Description

I have found many sources are missing from `generated-sources.json` when using the npm provider (under certain conditions). When it comes to running the flatpak-builder you see something like this:

```log
...
npm ERR! code ENOTCACHED
npm ERR! request to https://registry.npmjs.org/yauzl/-/yauzl-2.10.0.tgz failed: cache mode is 'only-if-cached' but no cached response is available.
```

I have a `package-lock.json` with over 1000 packages but `flatpak-node-generator` only produces 90 sources.

When I debugged the code of `flatpak-node-generator` I could see it was treating many packages' source as being of type `LocalSource` (which excludes them from generated sources) despite that not being the case. This is due (I believe) to a subtle bug in the code where if a package json file exists for a package in a local `node_modules` dir where `flatpak-node-generator` is run then the package will always be treated as local.

Deleting/renaming your `node_modules` dir before running `flatpak-node-generator` in the project's repository will workaround the issue but I am preparing a PR to fix the real issue also.

I believe this may also be the same bug as #244 which was closed when the OP switched to yarn.

Ian2020:
@hfiguiere this issue is not fixed by #427. When I use the suggested `--package-lock-only` flag to produce a package-lock.json any git URL dependencies (git+https) in the package.json are missing their own dependencies in the resulting `generated-sources.json`. This shows up as a `getaddrinfo` error during the flatpak build.

 Can this please be reopened and my PR below considered for merging? Thanks.

https://github.com/flatpak/flatpak-builder-tools/pull/378

razzeee:
@Ian2020 but your last comment in that issue seems, like you still wanted to hunt down the `file:` regression?

hfiguiere:
reopening for now.

Ian2020:
Thanks both! Yes I will continue with #378 next week to make it all good. Just wanted to check it was still def. under consideration as it's been a while 👍

PR #378, opened 2023-10-13
node: Correct the LocalSource check for npm provider

Ian2020:
Fix for #377.  The check was comparing a string against a PosixPath which always fails. The intention is to see if the package_json file has any parent called "node_modules". Here's an example of what I mean:

```python
@>>> from pathlib import Path
@>>> a = Path("./node_modules/ansi-gray")
@>>> "node_modules" in a.parents
False # WRONG
@>>> Path("node_modules") in a.parents
True # RIGHT
```

I wanted to write a test for this situation but was unsure how to structure it. I don't think the current tests in `test_providers.py` ever do a `node install` so there is no `node_modules` dir and thus this scenario is never tested. However I think it's more than likely to occur in reality.


node/flatpak_node_generator/providers/npm.py
```diff
@@ -110,7 +110,7 @@ def _process_packages_v2(
             source: PackageSource
             package_json_path = lockfile.parent / install_path / 'package.json'
             if (
-                'node_modules' not in package_json_path.parents
+                Path('node_modules') not in package_json_path.parents
```
refi64, line 113: OOF that's a nasty bug! If the intent is to see if *any* component is `node_modules` though, this also doesn't work:

```python3
>>> Path('node_modules') in Path('node_modules/x').parents
True
>>> Path('node_modules') in Path('a/node_modules/x').parents
False
>>>
```

Perhaps this should instead be:

```suggestion
                'node_modules' not in package_json_path.parts[:-1]
```

(not actually sure if the `:-1` is needed to trim off the last part; the OG code used `parents`, but I dunno how important excluding the final component is)

refi64 (commented):
> I wanted to write a test for this situation but was unsure how to structure it.

We actually do have an existing test for installing local packages: see `tests/data/local`. However, the `package-lock.v3.json` inside it uses `file:`:

```json
{
  "name": "@flatpak-node-generator-tests/local",
  "version": "1.0.0",
  "lockfileVersion": 3,
  "requires": true,
  "packages": {
    "": {
      "name": "@flatpak-node-generator-tests/local",
      "version": "1.0.0",
      "dependencies": {
        "subdir": "file:subdir"
      }
    },
    "node_modules/subdir": {
      "name": "@flatpak-node-generator-tests/subdir",
      "version": "1.0.0",
      "resolved": "file:subdir"
    }
  }
}
```

so it ends up just following the `file:` branch later on, and thus this particular part wasn't tested. I...don't actually know what scenario results in this branch being taken, could you share the package-lock.json that reproduced the bug? (cc @gasinvein you might remember more about this than I do)

node/flatpak_node_generator/providers/npm.py
Ian2020: Good catch - yes I'm not exactly sure if we're just trying to catch any dir called `node_modules` in the path or if it has to be the last/first. Maybe I'll see if there's a spec for this or something in the node docs.

Ian2020:
My package-lock.json is nothing special I don't think (attached) - it's just that for many of my packages the second clause of the LocalSource test is already true (`package_json_path.exists()`):

```python
[package-lock.json](https://github.com/flatpak/flatpak-builder-tools/files/12903897/package-lock.json)

            if (
                Path('node_modules') not in package_json_path.parents
                and package_json_path.exists()
            ):
```

...so thanks to the first clause already being defective most of my packages end up considered local.

node/flatpak_node_generator/providers/npm.py
gasinvein: Yeah this definitely should've been `package_json_path.parts`. My bad.

Ian2020: That works for me, I've made the change.

Ian2020:
> We actually do have an existing test for installing local packages...however, the package-lock.v3.json inside it uses file...so it ends up just following the file: branch later on, and thus this particular part wasn't tested. I...don't actually know what scenario results in this branch being taken

I've rethought my approach here - please take a look at the latest changes. I've got to a solution that's a bit simpler now and I think the existing tests cover all the scenarios adequately.

Your comment above was correct to question why this "LocalSource" branch is needed if the `file:` case is already handled in another branch. If I remove the LocalSource branch completely we get some test failures but only in tests that use `package-lock.v3.json`. Those failures (it turns out) are due to the fact that the code tries to process the "root" package as a dependency as well as all the "real" dependencies. To explain further here's part of a typical v3 package-lock:

```json
...
  "packages": {
    "": {
      "name": "@flatpak-node-generator-tests/name-as-dep",
      "version": "1.0.0",
      "dependencies": {
        "word-wrap": "^1.2.3"
      }
    },
    "node_modules/word-wrap": {
      "version": "1.2.3",
      "resolved": "https://registry.npmjs.org/word-wrap/-/word-wrap-1.2.3.tgz",
      "integrity": "sha512-Hz/mrNwitNRh/HUAtM/VT/5VH+ygD6DV7mYKZAtHOrbs8U7lvPS6xf7EJKMF0uW1KJCl0H701g3ZGus+muE5vQ==",
      "license": "MIT",
      "engines": {
        "node": ">=0.10.0"
      }
    }
  }
...
```

That first package with an empty key is added by npm for the "root" project:

> The root project is typically listed with a key of "", and all other packages are listed with their relative paths from the root project folder.

From [package-lock.json](https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json).

This root project entry causes problems for `_process_packages_v2` as it mistakenly processes it as if it were a real dependency. Previously the LocalSource branch was catching this dependency, misclassifying the root project as a LocalSource and thus it was ignored, by not by intention.

If instead we add code to ignore the root package explicitly all the tests pass and we can remove the LocalSource branch completely.

I think this is a better approach and more clearly deals with the issue.

refi64:
Ah, nice investigation! A few small notes:

- It's not entirely clear to me why the test was reverted?
- It would make the history a bit cleaner if the commits were squashed into two, e.g.:
  ```
  node: Correct the LocalSource check for npm
  node: Explicitly ignore the root project
  ```
  with the first containing the `Cleaner expression` commit as well. (Normally I'd just do a squash-on-merge or similar, but in this case there's *two* separate changes, so squashing on merge would collapse them into 1 which isn't quite what we want either.)

dbrgn:
Thanks @Ian2020 for the work on this, I'm also running into the problems in #377.

I applied the patch in this PR, but noticed that actual local dependencies (starting with `file:`) from a npm lockfile v3 cannot be handled anymore:

    NotImplementedError: Don't know how to handle package libs/eslint-myplugin in ../../package-lock.json

When I only apply the first commit ("Correct the LocalSource check for npm"), but not the second one, it seems to work. (I run into #382, but that's another issue.)

Ian2020:
Hey @dbrgn thanks for the feedback, I'll take a look when I get some time.

Ian2020:
> I applied the patch in this PR, but noticed that actual local dependencies (starting with `file:`) from a npm lockfile v3 cannot be handled anymore:
> 
> ```
> NotImplementedError: Don't know how to handle package libs/eslint-myplugin in ../../package-lock.json
> ```

@dbrgn could you give me a sample lockfile that's showing this problem? The tests are passing for me and there are some local `file:` type dependencies in a v3 lockfile already in the suite i.e. https://github.com/flatpak/flatpak-builder-tools/blob/master/node/tests/data/packages/local/package-lock.v3.json so not sure what I've missed here.

Ian2020:
A fix has arrived in master for this recently I believe: https://github.com/flatpak/flatpak-builder-tools/commit/aac65cf44cd4e008594a9d9ac1db08e2025067a6

Closing in favour of my changes to support gitrefs which also incorporates this fix in https://github.com/flatpak/flatpak-builder-tools/pull/382

PR #379, opened 2023-10-14
node: Bump aiohttp for python 3.12 compatibility 

bbhtt:
pipx install fails to compile aiohttp on Fedora 39 (and any distro shipping python 3.12) because of https://github.com/aio-libs/aiohttp/issues/7229

Draft because waiting for aiohttp release a stable build.

Future aiohttp versions won't support python 3.7 and it is also not supported anymore https://devguide.python.org/versions/, so I dropped it


bbhtt:
Hm I forgot about the workflow

bbhtt:
aiohttp 3.9.0 is now available with python 3.12, marking as ready.

bbhtt:
@refi64 can you have a look at this, please?

Issue #237, opened 2021-10-24
git+ssh make flathub bot stuck at downloading


catsout:
From: 
https://github.com/npm/arborist/blob/d9c603ede4fefb0bda1385e70b431b3adec810bf/lib/shrinkwrap.js#L704

Now npm will resolve any github url to `git+ssh`.  
But this will make flathub bot stuck at downloading:  
https://github.com/flathub/com.getmailspring.Mailspring/pull/2#issuecomment-947355892  
https://github.com/flathub/net.cozic.joplin_desktop/pull/18#issuecomment-901512443  

As https://github.com/flathub/flathub/issues/2466#issuecomment-901517504, using "ssh" URL like that is bound to fail on flathub: there is no ssh key setup.  

So, could we add a option to replace the url, or add some warn messages?

olof-nord:
In case someone comes across this later: the node flatpak-builder-tools only supports a subset of what npm does.
What should work is the following: (see https://github.com/flatpak/flatpak-builder-tools/blob/master/node/tests/data/packages/minimal-git/package.json)

```json
{
  "name": "@flatpak-node-generator-tests/minimal-git",
  "version": "1.0.0",
  "dependencies": {
    "nop": "https://git@github.com/supershabam/nop.git"
  }
}
```

Issue #182, opened 2021-01-28
flatpak-node-generator.py failed to deal with deps with direct git reference 


proletarius101:
## What is a dependency with direct git reference?

https://github.com/standardnotes/desktop/blob/3fbb58adf16a564d6773aae6cce273325f9eeba8/package.json#L12
````json

  "dependencies": {
    ...
    "decrypt": "github:standardnotes/decrypt#master",
    ...
  },
````
## What did I do?
````
python flatpak-node-generator.py npm ../standardnotes-desktop/package-lock.json --xdg-layout  -r
````
## The results

````
npm ERR! code ENOTCACHED
npm ERR! request to https://registry.npmjs.org/@standardnotes%2fsncrypto-web failed: cache mode is 'only-if-cached' but no cached response available.
````
A complete log could be found at https://flathub.org/builds/#/builders/35/builds/4478

## What I can suggest

`crypto-web` is a dependency of `github:standardnotes/decrypt#master`. I think it's failed to be included in the cache.

proletarius101:
## Mitigation

1. Generate separate `generated-sources.json`s for the git-referenced modules.
2. Replace `os.mkdir('some_dir')` with `os.makedirs('some_dir', exist_ok=True)` in all `generated-sources.json`

proletarius101:
Some interesting logs:
````

21 silly removeObsoleteDep removing decrypt@github:standardnotes/decrypt#5d487038d3568d180e7cf561d774e1b685985d62 from the tree as its been replaced by a newer version or is no longer required
22 silly removeObsoleteDep removing @standardnotes/sncrypto-web@1.2.9 from the tree as its been replaced by a newer version or is no longer required
23 silly removeObsoleteDep removing @standardnotes/sncrypto-common@1.2.9 from the tree as its been replaced by a newer version or is no longer required
24 silly removeObsoleteDep removing libsodium-wrappers@0.7.8 from the tree as its been replaced by a newer version or is no longer required
25 silly removeObsoleteDep removing libsodium@0.7.8 from the tree as its been replaced by a newer version or is no longer required
26 silly removeObsoleteDep removing @standardnotes/snjs@2.0.22 from the tree as its been replaced by a newer version or is no longer required
````

proletarius101:
[...]

PR #382, opened 2023-11-03
[node] support git sources in lockfile v2 format

### LINKED ISSUE #381: [node] NotImplementedError: Git sources in lockfile v2 format are not supported yet
### flatpak-builder version

1.3.3

Fedora 38

npm package-lock.json

### Source repository URL

_No response_

https://github.com/Ian2020/io.onlykey.OnlyKey-App/blob/main/io.onlykey.OnlyKey-App.yaml

### Description

As title and as intended at the moment the node tool does not support git sources for lockfile v2.

For example fragment of a lockfile below, the nw dependency causes the issue:

```json
{
  "name": "OnlyKey",
  "version": "5.5.0",
  "lockfileVersion": 3,
  "requires": true,
  "packages": {
    "": {
      "name": "OnlyKey",
      "version": "5.5.0",
      "license": "Apache-2.0",
      "dependencies": {
        "auto-launch": "^5.0.5",
        "js-sha256": "^0.9.0",
        "nw": "git+https://git@github.com/Ian2020/npm-installer.git#8e1575a4abf16901924a6146835334ddeaa17b14",
        "nw-autoupdater": "^1.1.11",
        "request": "^2.88.2",
        "sshpk": "^1.17.0"
      },
      "devDependencies": {
        "chai": "^4.3.7",
        "chai-as-promised": "^7.1.1",
        "eslint": "^8.33.0",
        "fs-jetpack": "^5.1.0",
        "gulp": "^4.0.2",
        "gulp-sourcemaps": "^3.0.0",
        "json": "^11.0.0",
        "mocha": "^10.2.0",
        "nw-dev": "^3.0.1",
        "q": "^1.5.1",
        "selenium-webdriver": "^4.8.0",
        "tree-kill": "^1.2.2",
        "yargs": "^17.6.2"
      },
      "optionalDependencies": {
        "appdmg": "^0.6.4"
      }
    },
   ...
```

Ian2020:
Fixes #381 and I hope will help #237 and #182.

I've added a new set of tests to cover more types of git references: https and ssh both with a commit hash and without. The tests cover all versions of the lockfile format v1,2,3.

In my testing I found that it doesn't appear to be necessary to patch package-lock.json (npm.py:444), so that has been removed. If there's another reason that is needed let me know but the tests all pass without it.


refi64:
I think this has some overlap with #351 :sweat_smile: see the discussion there.

jwillikers:
Attempting to use this for the Stretchly flatpak, and I get the following error when building:

```
Downloading sources
Fetching git repo https://git@github.com/Jelmerro/dbus-final.git, ref refs/heads/master
remote: Total 0 (delta 0), reused 0 (delta 0), pack-reused 0
Fetching git repo https://github.com/hovancik/stretchly.git, ref refs/tags/v1.15.0
remote: Total 0 (delta 0), reused 0 (delta 0), pack-reused 0
Starting build of net.hovancik.Stretchly
Cache hit for dbus-glib, skipping build
Cache miss, checking out last cache hit
========================================================================
Building module stretchly in /home/jordan/Projects/net.hovancik.Stretchly/.flatpak-builder/build/stretchly-11
========================================================================
Note: switching to '3e43f60d80bbcdf0dfa0b59b838097d6af4d17ba'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at 3e43f60 update deps
jq: error: Could not open file /var/home/jordan/Projects/net.hovancik.Stretchly/.flatpak-builder/build/stretchly-11/package.json: No such file or directory
/var/home/jordan/Projects/net.hovancik.Stretchly/.flatpak-builder/build/stretchly-11/flatpak-node/patch/app.sh: line 2: /var/home/jordan/Projects/net.hovancik.Stretchly/.flatpak-builder/build/stretchly-11/app/package.json.new: No such file or directory
mv: cannot stat '/var/home/jordan/Projects/net.hovancik.Stretchly/.flatpak-builder/build/stretchly-11/app/package.json.new': No such file or directory
Error: module stretchly: Child process exited with code 1
```

Ian2020:
There's something up with the paths here:

```
Building module stretchly in /home/jordan/Projects/net.hovancik.Stretchly/.flatpak-builder/build/stretchly-11
```

...but later the jq is looking in `/var/home/...` rather than `/home/...`

```
jq: error: Could not open file /var/home/jordan/Projects/net.hovancik.Stretchly/.flatpak-builder/build/stretchly-11/package.json: No such file or directory
```

Not sure where that `/var` prefix has come from.

jwillikers:
> There's something up with the paths here:
> 
> ```
> Building module stretchly in /home/jordan/Projects/net.hovancik.Stretchly/.flatpak-builder/build/stretchly-11
> ```
> 
> ...but later the jq is looking in `/var/home/...` rather than `/home/...`
> 
> ```
> jq: error: Could not open file /var/home/jordan/Projects/net.hovancik.Stretchly/.flatpak-builder/build/stretchly-11/package.json: No such file or directory
> ```
> 
> Not sure where that `/var` prefix has come from.

That's a thing on Fedora rpm-ostree systems. `/home` is a symlink to `/var/home`.

jwillikers:
The manifest and everything can be found in [this MR](https://github.com/flathub/net.hovancik.Stretchly/pull/8).

Ian2020:
Thanks, I'll take a look.

Ian2020:
Ok a few things going on here:

* The inclusion of `generated-sources.json` in [net.hovancik.Stretchly.yaml](https://github.com/flathub/net.hovancik.Stretchly/blob/master/net.hovancik.Stretchly.yaml) needs to come AFTER the stretchly source code is checked out, that's why package.json can't be found to be patched by generated-sources. This should fix the error you've reported but you'll next hit a ENOTCACHED error because...
* We don't patch refs that looks like "github:user/project" and in Stretchly you have exactly this: "github:Jelmerro/dbus-final". I can't find in the [package.json spec](https://docs.npmjs.com/cli/v9/configuring-npm/package-json) where this is allowed but it seems to be common searching across Github. So I think I have to add support for that to this PR.
* The Stretchly project has two `package.json` files: one in `/` one in `/app`. Both would have to be patched for its flatpak build to succeed with git refs. The `electron-builder` cmd in the postinstall script seems to use the `/app` one. @jwillikers do you know why there are two `package.json` files?

jwillikers:
> Ok a few things going on here:
> 
>     * The inclusion of `generated-sources.json` in [net.hovancik.Stretchly.yaml](https://github.com/flathub/net.hovancik.Stretchly/blob/master/net.hovancik.Stretchly.yaml) needs to come AFTER the stretchly source code is checked out, that's why package.json can't be found to be patched by generated-sources. This should fix the error you've reported but you'll next hit a ENOTCACHED error because...
> 
>     * We don't patch refs that looks like "github:user/project" and in Stretchly you have exactly this: "github:Jelmerro/dbus-final". I can't find in the [package.json spec](https://docs.npmjs.com/cli/v9/configuring-npm/package-json) where this is allowed but it seems to be common searching across Github. So I think I have to add support for that to this PR.
> 
>     * The Stretchly project has two `package.json` files: one in `/` one in `/app`. Both would have to be patched for its flatpak build to succeed with git refs. The `electron-builder` cmd in the postinstall script seems to use the `/app` one. @jwillikers do you know why there are two `package.json` files?

I've changed that order.
As for why there are two `package.json` files, I'm not sure.
@hovancik Could you shed some light on why there is `app/package.json` in addition to `package.json`?

hovancik:
@jwillikers that's the structure from `electron-builder`: https://www.electron.build/tutorials/two-package-structure.html

I've tried in the past to remove 2 package structure, as they say it's not needed anymore but it did not work for me.

Ian2020:
> * We don't patch refs that looks like "github:user/project" and in Stretchly you have exactly this: "github:Jelmerro/dbus-final". I can't find in the [package.json spec](https://docs.npmjs.com/cli/v9/configuring-npm/package-json) where this is allowed but it seems to be common searching across Github. So I think I have to add support for that to this PR.

Ok I've added support for these 'shortcut' git urls to the PR. But we still need to overcome the 2 package structure in stretchly somehow.

hovancik:
@Ian2020 should I try to remove 2package structure? It's been some time since last time I tried so I can try again if that helps

Ian2020:
@hovancik yes I think that would be simplest if you can, thanks.

hovancik:
@Ian2020 done, new version is out.

hovancik:
@Ian2020 actually scratch that, it does not work, will need to look into this more.

Ian2020:
Ok, despite that I can now successfully build Stretchly v1.15.1 as a flatpak. I created `generated-sources.yaml` using this PR against v1.15.1 of stretchly. Then copied that file into branch `update-71b455c` of https://github.com/flathub/net.hovancik.Stretchly and updated the source tag in the yaml to v1.15.1 and it builds a flatpak that I could run @hovancik @jwillikers

So I think this PR is doing what it should do for Stretchly to build as a flatpak, I hope you can sort the other issues in Stretchly. Let me know if you think there's anything more needed here.

jwillikers:
@Ian2020 One thing, I noticed that running `flatpak-node-generator` from this PR doesn't work when attempting to run it from a relative directory. I usually run `flatpak-node-generator npm stretchly/package-lock.json --electron-node-headers` with Stretchly checked out in the stretchly subdirectory in the Flatpak repository. I had to change into the stretchly subdirectory, generate the sources, and then move the generated file up one directory.

hovancik:
@Ian2020 can you possibly try with this PR? https://github.com/hovancik/stretchly/pull/1401 

Seems like it was local issue only but would be nice to confirm build is ok after this change

jwillikers:
> @Ian2020 can you possibly try with this PR? [hovancik/stretchly#1401](https://github.com/hovancik/stretchly/pull/1401)
> 
> Seems like it was local issue only but would be nice to confirm build is ok after this change

@hovancik  I just tested it and that commit builds and runs fine.

hovancik:
Thanks, we are all good then, I guess.

hfiguiere:
I don't maintain that module and I don't have the knowledge space.

node/flatpak_node_generator/providers/__init__.py
```diff
@@ -33,6 +34,14 @@ def parse_git_source(self, version: str, from_: Optional[str] = None) -> GitSour
         if not new_url.netloc:
             path = new_url.path.split('/')
             new_url = new_url._replace(netloc=path[0], path='/'.join(path[1:]))
+        # Replace https://git@github.com:ianstormtaylor/to-camel-case.git
+        # wth     https://git@github.com/ianstormtaylor/to-camel-case.git
```
dbrgn, line 38: Typo:

```suggestion
        # with    https://git@github.com/ianstormtaylor/to-camel-case.git
```

node/flatpak_node_generator/providers/npm.py
```diff
@@ -425,72 +422,79 @@ def _finalize(self) -> None:
...
-                """,
-            }
+            # Generate jq script to patch the package.json file
+            script = r"""
+                        walk(
+                            if type == "object"
+                            then
+                                to_entries | map(
+                                    if (.value | type == "string") and $data[.value]
+                                    then .value = "git+file:\($buildroot)/\($data[.value])"
```
jwillikers, line 432: Using this fork to update the dependencies for the Stretchly Flatpak resulted in the following error:

```
npm error code 128
npm error An unknown git error occurred
npm error command git --no-replace-objects ls-remote file:/var/home/jordan/Projects/net.hovancik.Stretchly/.flatpak-builder/build/stretchly-14/flatpak-node/git-packages/dbus-final-3e43f60d80bbcdf0dfa0b59b838097d6af4d17ba
npm error ssh: Could not resolve hostname file: Temporary failure in name resolution
npm error fatal: Could not read from remote repository.
npm error
npm error Please make sure you have the correct access rights
npm error and the repository exists.
```

I was able to fix this by using the syntax `file://` instead of just `file:`.
This matches the documented syntax for [Git URL's](https://git-scm.com/docs/git-fetch#_git_urls).

```suggestion
                                    then .value = "git+file://\($buildroot)/\($data[.value])"
```

jwillikers:
This doesn't work when the given a branch name instead of a commit hash in the `package.json` file:

```
"windows-quiet-hours": "github:hovancik/windows-quiet-hours#fix-LPTSTR"
```

This occurs for the Stretchly Flatpak. It fails to patch this reference in the  `package.json` file to point to the location of the source in the cache.

node/flatpak_node_generator/providers/npm.py
jwillikers: I think this is only an issue for node18.

Ian2020: I've just hit this problem myself with another flatpak so will get my head back into this PR (and the problem with branches below) and hope to get it polished up for merging.

hadess:
Running the latest version of the patch against this [lockfile](https://github.com/FelicianoTech/pocket-casts-desktop-app/blob/trunk/package-lock.json) throws this error at build time:
```
⠙⠹npm warn skipping integrity check for git dependency ssh://git@github.com/electron/node-gyp.git
⠹⠸⠼⠴npm warn deprecated gar@1.0.4: Package no longer supported. Contact Support at https://www.npmjs.com/support for more info.
⠴npm warn deprecated @npmcli/move-file@2.0.1: This functionality has been moved to @npmcli/fs
⠦npm warn deprecated rimraf@2.6.3: Rimraf versions prior to v4 are no longer supported
⠦npm warn deprecated rimraf@3.0.2: Rimraf versions prior to v4 are no longer supported
⠦npm warn deprecated glob@8.1.0: Glob versions prior to v9 are no longer supported
⠦npm warn deprecated glob@8.1.0: Glob versions prior to v9 are no longer supported
⠦npm warn deprecated sudo-prompt@9.2.1: Package no longer supported. Contact Support at https://www.npmjs.com/support for more info.
⠦⠧⠇⠏⠋⠙⠹npm error code ENOTCACHED
⠹npm error request to https://codeload.github.com/electron/node-gyp/tar.gz/06b29aafb7708acef8b3669835c8a7857ebc92d2 failed: cache mode is 'only-if-cached' but no cached response is available.
```

Looks like the `node-gyp` module got generated as a `git` checkout, not a `tar.gz` download:
```
    {    
        "type": "git",
        "url": "https://git@github.com/electron/node-gyp.git",
        "commit": "06b29aafb7708acef8b3669835c8a7857ebc92d2",
        "dest": "flatpak-node/git-packages/@electron/node-gyp-06b29aafb7708acef8b3669835c8a7857ebc92d2"
    },   
```

See https://github.com/flathub/tech.feliciano.pocket-casts/pull/11

node/flatpak_node_generator/providers/npm.py
Ian2020: This suggested change fixes the node18 flatpak I'm working on so I've committed it. My flatpak was moved from npm v9 to v10 when org.freedesktop.Sdk.Extension.node18 was updated in May 2024. I think this addresses an issue with how npm v10 is working with these refs.

Ian2020:
@jwillikers please try again with today's commits to this branch. I've added tests and a fix for your scenario, should work for deps with a branch now I hope.

Ian2020:
Today's commit addresses all issues raised so far. I can build both Stretchly and Pocket-Casts flatpaks no problem. Please let me know of any new problems. About these latest changes:

- Added new tests and fixed previous tests to be more realistic - the lockfiles were hand-edited before but I regenerated them all with the right version of npm and yarn. This surfaced that we **do** need to patch package.json and package-lock.json but **only for lockfile v1/npm v7**. If support can ever be dropped for this old format lots of code can go. As such that code is all protected by a guard `if lockfile_v1` in `npm.py` so it's clear what can be dropped in future.
- Due to above we only add patch commands to the `generated-sources.json` when needed now - most projects which are NPM-16+ do not need them.
- @hadess discovered an issue with transitive git dependencies in their flatpak. New tests have been added to cover this. Another entry was needed in the npm cache essentially. We also found a separate issue with electron forge: https://github.com/flatpak/flatpak-builder-tools/issues/452
- Support for git refs in lockfile v1 files has been completed here also. This should have eliminated all "Don't know how to handle package" and "KeyError: 'integrity'" exceptions I hope. I found a few open bugs for those in this repo and have commented that this PR should help.

hadess:
This is blocking updating pocket-casts:
https://github.com/flathub/tech.feliciano.pocket-casts/pull/21
as @Ian2020's repo is outside the allowed organisations:
```
        "external-gitmodule-url-found: Only flatpak, flathub, and flathub-infra gitmodules are allowed in manifest git repo: ['https://github.com/Ian2020/flatpak-builder-tools.git']"
```

And it fixes issues that otherwise cause dependency generation not to work. Using Ian2020's repo didn't cause problems 2 weeks ago (https://github.com/flathub/tech.feliciano.pocket-casts/pull/8#issuecomment-2816669606), but it now does.

Let me know if I should file an issue about this (which I definitely will if this cannot not be fixed quickly).

(Note to self, the repo to file it against is https://github.com/flathub-infra/vorarbeiter/)

salim-b:
@hadess Instead of [using Ian2020's fork](https://github.com/flathub/tech.feliciano.pocket-casts/blob/master/.gitmodules), you could just reference this PR. To e.g. install it via pipx:

```sh
pipx install --force git+https://github.com/flatpak/flatpak-builder-tools.git@refs/pull/382/head#subdirectory=node
```

bbhtt:
External gitmodule check was added to linter several months ago and it doesn't have anything to do with vorarbeither.

bbhtt:
Is the work here still ongoing? If not if you can fix the conflict, rebase and clear up the history it'd be nice.

> I think this has some overlap with https://github.com/flatpak/flatpak-builder-tools/pull/351 😅 see the discussion there.

@refi64 Seems like this has progressed a lot more than the other PR. Any chance you can review it?

hadess:
> @hadess Instead of [using Ian2020's fork](https://github.com/flathub/tech.feliciano.pocket-casts/blob/master/.gitmodules), you could just reference this PR. To e.g. install it via pipx:
> 
> ```shell
> pipx install --force git+https://github.com/flatpak/flatpak-builder-tools.git@refs/pull/382/head#subdirectory=node
> ```

This worked (and I learnt something), thanks!

Ian2020:
> if you can fix the conflict, rebase and clear up the history it'd be nice.

@bbhtt can do, probably be next week now

Ian2020:
> if you can fix the conflict, rebase and clear up the history it'd be nice.

Done.

node/flatpak_node_generator/providers/npm.py
```diff
@@ -459,56 +497,57 @@ def _finalize(self) -> None:
...
+                patch_dest = (
+                    self.gen.data_root / 'patch' / self.relative_lockfile_dir(lockfile)
+                )
+                # Don't use with_extension to avoid problems if the package has a . in its name.
+                patch_dest = patch_dest.with_name(patch_dest.name + '.sh')
 
-            self.gen.add_script_source(patch_commands[lockfile], patch_dest)
-            patch_all_commands.append(f'"$FLATPAK_BUILDER_BUILDDIR"/{patch_dest}')
+                self.gen.add_script_source(patch_commands[lockfile], patch_dest)
+                patch_all_commands.append(f'$FLATPAK_BUILDER_BUILDDIR/{patch_dest}')
```
bbhtt, line 544: why was the quote removed? That came from a PR to solve a bug when the pwd has spaces in it, same for below

node/flatpak_node_generator/providers/npm.py
```diff
@@ -369,9 +367,49 @@ async def generate_package(self, package: Package) -> None:
             # Get a unique name to use for the Git repository folder.
             name = f'{package.name}-{source.commit}'
             path = self.gen.data_root / 'git-packages' / name
-            self.git_sources[package.lockfile][path] = source
+            self.git_sources[package.lockfile][path] = (package.name, source)
             self.gen.add_git_source(source.url, source.commit, path)
 
+            url = urllib.parse.urlparse(source.url)
```
bbhtt, line 373: is the source here always going to github.com and gitlab.com?

node/flatpak_node_generator/providers/npm.py
Ian2020: I didn't notice that change when I rebased, added both back now.

node/flatpak_node_generator/providers/npm.py
Ian2020: When I was working on this github and gitlab were the only two forges on which I could easily find packages to test, however npm seems to also support bitbucket and gists from their docs: https://docs.npmjs.com/cli/v11/commands/npm-install.

I agree the code should be fixed though as any other url will fail. Is it ok to throw an exception if it's not a url we can handle? My suspicion is that gitlab/github covers 80%+ of use cases. If anyone can provide a failing url it could be addressed in future as we need to determine where to get the tarball from and how npm wants it indexed.

bbhtt: > s it ok to throw an exception if it's not a url we can handle?

Ok let's do that for now.

node/flatpak_node_generator/providers/__init__.py
```diff
@@ -33,6 +34,14 @@ def parse_git_source(self, version: str, from_: Optional[str] = None) -> GitSour
         if not new_url.netloc:
             path = new_url.path.split('/')
             new_url = new_url._replace(netloc=path[0], path='/'.join(path[1:]))
+        # Replace https://git@github.com:ianstormtaylor/to-camel-case.git
+        # with    https://git@github.com/ianstormtaylor/to-camel-case.git
+        # for git+ssh URLs
+        if ':' in new_url.netloc:
+            netloc_split = new_url.netloc.split(':')
```
refi64, line 41: this will break if there are extra colons (I mean, there really *shouldn't* be, but alas)
```suggestion
            netloc_split = new_url.netloc.split(':', 1)
```

(note that the branch immediately above also handles this, but...it does so kinda poorly, by comparison)

node/flatpak_node_generator/providers/npm.py
```diff
@@ -79,11 +80,17 @@ def _process_packages_v1(
...
-                integrity = Integrity.parse(info['integrity'])
+                integrity = None
+                if 'integrity' in info:
+                    integrity = Integrity.parse(info['integrity'])
+
                 if 'resolved' in info:
-                    source = ResolvedSource(
-                        resolved=info['resolved'], integrity=integrity
-                    )
+                    if info['resolved'].startswith('git+'):
```
refi64, line 88: `info['resolved']` should probably be in an intermediate variable so the same string constant isn't copy-pasted in 3 places. (this is also something the rest of the code admittedly does poorly.)

node/flatpak_node_generator/providers/npm.py
refi64: I believe the full list of stuff to handle would be in: https://github.com/npm/hosted-git-info

node/tests/data/packages/git/package-lock.v1.json
```diff
@@ -0,0 +1,94 @@
+{
```
refi64, line 1: very nice to see the tests for this!

node/tests/test_providers.py
```diff
@@ -30,6 +27,34 @@ async def test_minimal_git(
...
+                {
+                    'type': 'dir',
+                    'path': str(shared_datadir / 'packages' / 'git' / 'subdir'),
+                    'dest': 'subdir',
+                }
+            ],
+        ),
+        commands=[
+            provider_factory_spec.install_command,
+            """node -e 'require("array-range");require("is-empty-object");require("is-number");require("person-lib");require("to-camel-case");require("to-capital-case");require("to-no-case");require("to-space-case");require("@flatpak-node-generator-tests/subdir").sayHello()'""",
```
refi64, line 52: this is...a bit horrifying, you could pull out the JS into a top-level multiline string constant, or just stick it in the shared datadir with the `package.json` (we don't really do that anywhere else, but the other lines are also *much* shorter!).

node/flatpak_node_generator/providers/npm.py
```diff
@@ -104,20 +111,15 @@ def _process_packages_v2(
...
                 continue
+            if install_path == '':
+                # The root project is typically listed with a key of ''
+                continue
 
             name = info.get('name')
 
             source: PackageSource
-            package_json_path = lockfile.parent / install_path / 'package.json'
-            if (
```
refi64, line 112: what's the reason behind this change? if "entry without a `resolved` is always a local path" is reliable, then yeah it would make removing this hack doable...but that would ideally be in a separate commit before this one, to be able to tell the logic apart more easily.

node/flatpak_node_generator/providers/npm.py
```diff
@@ -369,9 +367,49 @@ async def generate_package(self, package: Package) -> None:
...
             name = f'{package.name}-{source.commit}'
             path = self.gen.data_root / 'git-packages' / name
-            self.git_sources[package.lockfile][path] = source
+            self.git_sources[package.lockfile][path] = (package.name, source)
             self.gen.add_git_source(source.url, source.commit, path)
 
+            url = urllib.parse.urlparse(source.url)
+            if url.netloc == 'git@github.com' or url.netloc == 'github.com':
+                url = url._replace(
+                    netloc='codeload.github.com', path=re.sub('.git$', '', url.path)
```
refi64, line 376: should the `.` in `.git` here and below be escaped? might be useful to extract this into a separate `git_suffix = re.compile(r'\.git$')` and then use that for the substitutions below.

node/flatpak_node_generator/providers/npm.py
```diff
@@ -369,9 +367,49 @@ async def generate_package(self, package: Package) -> None:
             # Get a unique name to use for the Git repository folder.
             name = f'{package.name}-{source.commit}'
             path = self.gen.data_root / 'git-packages' / name
-            self.git_sources[package.lockfile][path] = source
+            self.git_sources[package.lockfile][path] = (package.name, source)
             self.gen.add_git_source(source.url, source.commit, path)
 
+            url = urllib.parse.urlparse(source.url)
+            if url.netloc == 'git@github.com' or url.netloc == 'github.com':
```
refi64, line 374: could be `url.hostname == 'github.com'` (and maybe also check `url.username in ('git', '')` or similar? do you know if that's semantically important in any way? if not, then just checking the hostname is a bit cleaner.) ditto for the `gitlab.com` check below.

node/flatpak_node_generator/providers/npm.py
```diff
@@ -208,7 +206,7 @@ def __init__(
         self.all_lockfiles: Set[Path] = set()
         # Mapping of lockfiles to a dict of the Git source target paths and GitSource objects.
         self.git_sources: DefaultDict[
-            Path, Dict[Path, GitSource]
+            Path, Dict[Path, Tuple[str, GitSource]]
```
refi64, line 209: should update the comment above to indicate what this is, since it's no longer *just* a GitSource. or maybe the value a dataclass like `NamedGitSource` or something instead, to make it clearer what the `str` is (and then you can just use attributes to get the innards, instead of having to do unpacking assignments)

node/flatpak_node_generator/providers/npm.py
```diff
@@ -459,56 +497,57 @@ def _finalize(self) -> None:
...
-
-        patch_all_commands: List[str] = []
-        for lockfile in self.all_lockfiles:
-            patch_dest = (
-                self.gen.data_root / 'patch' / self.relative_lockfile_dir(lockfile)
-            )
-            # Don't use with_extension to avoid problems if the package has a . in its name.
-            patch_dest = patch_dest.with_name(patch_dest.name + '.sh')
+                with open(lockfile) as fp:
+                    lockfile_v1 = json.load(fp)['lockfileVersion'] == 1
```
refi64, line 501: this could probably be saved inside `self.git_sources`' values, right? and then no need to re-open the lockfile here.

node/flatpak_node_generator/providers/npm.py
Ian2020: Fixed, thanks.

node/tests/data/packages/git/package-lock.v1.json
Ian2020: Thanks, getting all the tests cases across all the various npm versions and lockfiles was quite an ordeal!

node/tests/test_providers.py
Ian2020: Agreed, moved to a top-level constant.

node/flatpak_node_generator/providers/npm.py
Ian2020: Agreed, when we've worked through all other suggestions to satisfaction I'll pull this out to its own commit and rebase the branch again ready for final merge.

node/flatpak_node_generator/providers/npm.py
Ian2020: Done, thanks.

node/flatpak_node_generator/providers/npm.py
```diff
@@ -369,9 +367,49 @@ async def generate_package(self, package: Package) -> None:
...
+            self.git_sources[package.lockfile][path] = (package.name, source)
             self.gen.add_git_source(source.url, source.commit, path)
 
+            url = urllib.parse.urlparse(source.url)
+            if url.netloc == 'git@github.com' or url.netloc == 'github.com':
+                url = url._replace(
+                    netloc='codeload.github.com', path=re.sub('.git$', '', url.path)
+                )
+                tarball_url = url._replace(
+                    path=re.sub('$', f'/tar.gz/{source.commit}', url.path)
```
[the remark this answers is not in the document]
Ian2020: Ha, I went regex crazy here. That's all fixed and simplified now to just use concatenation.

node/flatpak_node_generator/providers/npm.py
Ian2020: After some thought I decided the lockfile version should not belong in git_sources as the lockfile version is per lockfile, not per git source. So in a separate commit ("don't read lockfile version twice) I've created another new dataclass 'Lockfile' to store the lockfile path and read the version when we first see it. I think this new dataclass fits in well with the other dataclasses in `package.py`.

One possible issue with this is in other places in the code we have variables `lockfile : Path` which treat the lockfile as just a path. So you might think this confuses things, let me know if there is a better approach.

A side-effect of this is that the yarn code path now also uses the new dataclass, so has access to its lockfile version too. It's hard-coded to '1' right now and unused, but might make adding support for v2 slightly clearer in future.

bbhtt:
@refi64 do you want to have another look at this?

node/flatpak_node_generator/providers/npm.py
refi64: oh I really like this overall, yeah not used everywhere but it's a good way to future-proof a bit

refi64 (commented):
LGTM now, thanks for this

PR #387, opened 2024-01-21
JSON to YAML instructions

Kentoseth:
how to convert json to yaml


TingPing:
All usage of this repo implies cloning it, so I'd rephrase it a bit to just include the command to run. (And then its executable so you can remove `python3`)

Kentoseth:
> All usage of this repo implies cloning it, so I'd rephrase it a bit to just include the command to run. (And then its executable so you can remove `python3`)

Let me know if I got the right executable if the repo is cloned.

PR #390, opened 2024-01-31
cargo: Fix KeyError for workspaces that don't contain all sections

felinira:
I stumbled upon this when trying to build gtk-rs-core.

https://github.com/gtk-rs/gtk-rs-core/blob/master/Cargo.toml

The workspace itself doesn't have a dev-dependencies key, so it will throw a KeyError.

```
  File "flatpak-cargo-generator.py", line 139, in normalized
    package[section_key][key] = self.workspace[section_key][key]
                                ~~~~~~~~~~~~~~^^^^^^^^^^^^^
KeyError: 'dev-dependencies'
```


hfiguiere:
Closed with #392

PR #392, opened 2024-02-09
Fix cargo workspace replacement for git deps

tarkah:
Current implementation assumes all workspace dependencies in workspace crates are single entry maps of
`<dep>.workspace = true` and overrides the `<dep>` value from the workspace

This fails if the workspace crate includes other keys for the `<dep>`, such as `optional = true` or
`features = []`, since those are dropped.

The current implementation also failed to properly recurse to capture `target.cfg(..).dependencies`.

This has been updated to properly fix the above

This fix was used to successfully generate sources for and build https://github.com/flathub/org.squidowl.halloy/pull/5

### DIFF
--- a/cargo/flatpak-cargo-generator.py
+++ b/cargo/flatpak-cargo-generator.py
@@ -116,6 +116,31 @@ def fetch_git_repo(git_url: str, commit: str) -> str:
         subprocess.run(['git', 'checkout', commit], cwd=clone_dir, check=True)
     return clone_dir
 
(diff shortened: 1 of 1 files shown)


tarkah:
This should also resolve https://github.com/flatpak/flatpak-builder-tools/pull/390

hfiguiere:
Isn't this a dupe of #390 ?

hfiguiere:
So this close #390 ?

hfiguiere:
Close #390

PR #393, opened 2024-02-11
Skip directory sources in Poetry lockfiles

Newbytee:
Directory sources, or "path dependencies" as the Poetry documentation[1] calls them, are by definition not available on PyPi. As such, there is no point in attempting to download these since there is no way to tell where the source could be acquired from the lockfile alone.

 [1]: https://python-poetry.org/docs/dependency-specification/#path-dependencies


hfiguiere:
Look sensible to me.

PR #395, opened 2024-02-11
Add Swift Package Manager (SPM) support

david-swift:
Hi!

This PR adds support for [SPM](https://www.swift.org/documentation/package-manager/) (e.g. useful in combination with the [Swift 5 SDK Extension](https://github.com/flathub/flathub/pull/4935)). It contains a script for generating the necessary  files, a README and a sample Swift package.

### DIFF
--- a/spm/README.md
+++ b/spm/README.md
@@ -31,8 +31,7 @@ The tool creates a `generated-sources.json` and a `setup-offline.sh` file in the
     "build-commands": [
         "./setup-offline.sh",
         "swift build -c release --static-swift-stdlib",
-        "mkdir /app/bin",
-        "cp .build/release/quickstart /app/bin"
+        "install -Dm755 .build/release/quickstart /app/bin/quickstart"
     ]
 }
 ```
--- a/spm/README.md
+++ b/spm/README.md
@@ -29,7 +29,6 @@ The tool creates a `generated-sources.json` and a `setup-offline.sh` file in the
         "generated-sources.json"
     ],
     "build-commands": [
-        "chmod +x ./setup-offline.sh ./setup-offline.sh",
         "./setup-offline.sh",
         "swift build -c release --static-swift-stdlib",
         "rm -r -f /app/bin",
--- a/spm/README.md
+++ b/spm/README.md
@@ -1,0 +1,43 @@
+# flatpak-spm-generator
+
+Tool to automatically generate `flatpak-builder` manifest JSON from a Swift package.
+
+## Requirements
+
+Swift is required to execute the script. It is not only written as a Swift script, but also uses Swift to get information about the dependencies of the package.
+
+## Usage
+
+The first step is to convert the dependencies declared in the `Package.swift` file as well as the dependencies of those dependencies into a format flatpak-builder can understand,
+and generate a script that makes SPM use the offline dependencies.
+```
+swift flatpak-spm-generator.swift ./quickstart ./quickstart
+```
+
+The first path is the directory containing the Swift package, and the second path is the directory containing the Flatpak manifest file.
+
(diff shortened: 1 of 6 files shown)


spm/README.md
```diff
@@ -0,0 +1,43 @@
...
+    "buildsystem": "simple",
+    "sources": [
+        {
+            "type": "dir",
+            "path": "."
+        },
+        "generated-sources.json"
+    ],
+    "build-commands": [
+        "chmod +x ./setup-offline.sh ./setup-offline.sh",
```
TingPing, line 32: Just make it executable at time of creation?

spm/README.md
```diff
@@ -0,0 +1,43 @@
...
+        "./setup-offline.sh",
+        "swift build -c release --static-swift-stdlib",
+        "rm -r -f /app/bin",
```
TingPing, line 35: Why is this necessary?

TingPing (commented):
Thanks for the contribution, Swift support would be great!

Are you going to commit to maintaining this here? As-in, is it OK to set you as a codeowner so new issues are assigned to you.

spm/README.md
david-swift: It actually isn't...

david-swift:
> Thanks for the contribution, Swift support would be great!
> 
> Are you going to commit to maintaining this here? As-in, is it OK to set you as a codeowner so new issues are assigned to you.

@TingPing  Yes, that's ok. Thanks for the review!

spm/README.md
```diff
@@ -0,0 +1,41 @@
...
+            "type": "dir",
+            "path": "."
+        },
+        "generated-sources.json"
+    ],
+    "build-commands": [
+        "./setup-offline.sh",
+        "swift build -c release --static-swift-stdlib",
+        "mkdir /app/bin",
+        "cp .build/release/quickstart /app/bin"
```
TingPing, line 35:
```suggestion
        "install -Dm755 .build/release/quickstart /app/bin/quickstart",
```

spm/quickstart/Sources/quickstart/main.swift
```diff
@@ -0,0 +1,26 @@
+// The Swift Programming Language
+// https://docs.swift.org/swift-book
+
+import Adwaita
+
+@main
+struct QuickStart: App {
+
+    let id = "com.flatpak.quickstart"
```
hfiguiere, line 9:
```suggestion
    let id = "org.flatpak.quickstart"
```

spm/quickstart/com.flatpak.quickstart.json
```diff
@@ -0,0 +1,39 @@
+{
+    "app-id": "com.flatpak.quickstart",
```
hfiguiere, line 2:
```suggestion
    "app-id": "org.flatpak.quickstart",
```

and the file should be renamed.

I know it's just an example, but at least it should conform to the requirements.

hfiguiere:
My biggest issue here is that the tool is in Swift. Nothing against Swift but it makes it more difficult to run and to maintain when everything else is in Python.

david-swift:
> My biggest issue here is that the tool is in Swift. Nothing against Swift but it makes it more difficult to run and to maintain when everything else is in Python.

Rewriting in Python wouldn't be too difficult, but the generation itself relies on Swift being installed. Isn't the idea that the app maintainers run those tools themselves? Installing Swift in a GitHub action for example is really simple, and people using this tool probably have Swift installed on their machine. Or do I misunderstand something, @hfiguiere?

spm/quickstart/com.flatpak.quickstart.json
hfiguiere: oh, I should fix that one then.

TingPing:
I do agree being in Python would be a little easier to maintain and use.

However I'm fine merging this as-is. You're putting in the work and if Swift works best for you then that's fine.

hfiguiere:
fine by me too. I just wanted to note the caveats

PR #397, opened 2024-02-21
spm: Fixes

hfiguiere:
spm: Add the script to the generate json
 
Create the folders in the setup script.

The rationale is to only have to add `generated-sources.json` to the manifest, so we need to list the `setup-offline.sh` too.

The `setup-offline.sh` need to create the destination directories otherwise it fails.

spm: Fix the generation when a package name is the substring of another
    
Example: `SwiftPangoCairo` and `SwiftPango`. The latter would be skipped.

(trying to build https://github.com/flathub/co.luoja.Telyn from source)


david-swift (approved):
Thank you so much!

Issue #367, opened 2023-08-31
cargo generator cant seem to find `build-dependencies`


ImUrX:
### flatpak-builder version

1.2.3

### Linux distribution and version

NixOS 23.11.20230830.e7f38be

### Affected flatpak-builder tool

cargo/flatpak-cargo-generator.py

### Source repository URL

https://github.com/SlimeVR/SlimeVR-Server.git

### Flatpak-builder manifest URL

https://github.com/flathub/dev.slimevr.SlimeVR.git

### Description

I'm trying to generate a sources.json and it seems its trying to find `build-dependencies` item when it doesn't exist, I'm really unsure for the reason. I saw that I had a repeated dependency and fixed it but it still happens.

madoka773:
```
⬢[root@toolbox io.github.keiko233.clash-nyanpasu]# ../../../flatpak-builder-tools/cargo/flatpak-cargo-generator.py ../../../clash-rs/Cargo.lock -o clash-rs/clash-rscargo-sources.json
INFO:root:Loading packages from https://github.com/watfaq/boring
INFO:root:Adding package boring from https://github.com/watfaq/boring
INFO:root:Adding package boring-sys from https://github.com/watfaq/boring
INFO:root:Adding package hyper-boring from https://github.com/watfaq/boring
INFO:root:Loading packages from https://github.com/watfaq/netstack-lwip
INFO:root:Adding package netstack-lwip from https://github.com/watfaq/netstack-lwip
INFO:root:Adding package tokio-boring from https://github.com/watfaq/boring
INFO:root:Loading packages from https://github.com/watfaq/rust-tun
INFO:root:Adding package tun from https://github.com/watfaq/rust-tun
Traceback (most recent call last):
  File "/root/Projects/repos/manifests/io.github.keiko233.clash-nyanpasu/../../../flatpak-builder-tools/cargo/flatpak-cargo-generator.py", line 449, in <module>
    main()
  File "/root/Projects/repos/manifests/io.github.keiko233.clash-nyanpasu/../../../flatpak-builder-tools/cargo/flatpak-cargo-generator.py", line 442, in main
    generated_sources = asyncio.run(generate_sources(load_toml(args.cargo_lock),
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/asyncio/runners.py", line 194, in run
    return runner.run(main)
[...]

PR #399, opened 2024-02-29
Find all crates in git repositories, matching cargo's behavior

sgvictorino:
Right now, `get_git_repo_packages` only finds git dependencies that are:

- their repo's root crate (`Cargo.toml` or `*/Cargo.toml`)
- in the dependency or workspace member trees of those root crates

Otherwise, `get_git_package_sources` throws a KeyError.

I found this with `eframe`:

```toml
package.name = "uses-eframe"
dependencies.eframe.git = "https://github.com/emilk/egui"
```

```sh
cargo generate-lockfile # succeeds
flatpak-cargo-generator.py Cargo.lock -d # KeyError: 'eframe'
```

<details>
<summary>Full Debug Log</summary>
<pre>
<code>
DEBUG:asyncio:Using selector: EpollSelector
DEBUG:root:depends-on-eframe has no source
INFO:root:Loading packages from https://github.com/emilk/egui
DEBUG:root:Loading workspace member ./crates/ecolor/Cargo.toml in /home/me/.cache/flatpak-cargo/https_github.com_emilk_egui
DEBUG:root:Loading workspace member ./crates/egui_demo_app/Cargo.toml in /home/me/.cache/flatpak-cargo/https_github.com_emilk_egui
DEBUG:root:Loading workspace member ./crates/egui_demo_lib/Cargo.toml in /home/me/.cache/flatpak-cargo/https_github.com_emilk_egui
DEBUG:root:Loading workspace member ./crates/egui_extras/Cargo.toml in /home/me/.cache/flatpak-cargo/https_github.com_emilk_egui
DEBUG:root:Loading workspace member ./crates/egui_glow/Cargo.toml in /home/me/.cache/flatpak-cargo/https_github.com_emilk_egui
DEBUG:root:Loading workspace member ./crates/egui_plot/Cargo.toml in /home/me/.cache/flatpak-cargo/https_github.com_emilk_egui
DEBUG:root:Loading workspace member ./crates/egui-wgpu/Cargo.toml in /home/me/.cache/flatpak-cargo/https_github.com_emilk_egui
DEBUG:root:Loading workspace member ./crates/egui-winit/Cargo.toml in /home/me/.cache/flatpak-cargo/https_github.com_emilk_egui
DEBUG:root:Loading workspace member ./crates/egui/Cargo.toml in /home/me/.cache/flatpak-cargo/https_github.com_emilk_egui
[...]

PR #323, opened 2022-10-09
cargo: Handle excluded paths in workspaces

gasinvein:
Should fix #322

PR #405, opened 2024-04-05
Fixes to flatpak-cargo-generator

zecakeh:
Various errors I encountered while trying the script and then to build a Flatpak app.

See each commit message for details.

See the last commit of https://github.com/flathub/org.gnome.Fractal/pull/41 for the differences in the generated sources.

### DIFF
--- a/cargo/flatpak-cargo-generator.py
+++ b/cargo/flatpak-cargo-generator.py
@@ -123,8 +123,8 @@ def update_workspace_keys(pkg, workspace):
             for target in item.values():
                 update_workspace_keys(target, workspace)
             continue;
-        # dev-dependencies should reference root dependencies table from workspace
-        elif key == 'dev-dependencies':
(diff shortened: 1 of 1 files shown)


hfiguiere:
I wonder how this intersects with #323 and #325.

hfiguiere:
and also #399

hfiguiere:
And issue #367, seems that this would fix it.

zecakeh:
After checking all of those:

- There is no direct link/conflict with #323, although I don't recommend to merge it.
- There is no direct link/conflict with #325.
- The last commit of this PR is not necessary if #399 is merged (which I recommend over #323).
- It could possibly fix #367, if we knew what the actual error looks like, and if we could reproduce it.

zecakeh:
I rebased on master and dropped the last commit, to fix conflicts.

hfiguiere:
thanks!

PR #407, opened 2024-04-26
Simplify flatpak-cargo-generator.py dependency install with Poetry

pieterdd:
I like to give tools such as `flatpak-cargo-generator.py` a dedicated virtualenv with the necessary dependencies, and Poetry was an easy way to handle the logistics behind the scenes.

### DIFF
--- a/cargo/README.md
+++ b/cargo/README.md
@@ -2,18 +2,21 @@
 
 Tool to automatically generate `flatpak-builder` manifest json from a `Cargo.lock`.
 
-## Requirements:
+## Requirements
 
-Python 3.8+ with following modules:
+Poetry users can run `poetry install` and skip this.
 
+Otherwise install Python 3.8+ with these modules:
 - toml
 - aiohttp
 
 Generated manifests are supported by flatpak-builder 1.2.x or newer.
 
-## Usage:
+## Usage
 
-The first step is to convert the locked dependencies by Cargo into a format flatpak-builder can understand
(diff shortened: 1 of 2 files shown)


pieterdd:
Come to think of it: this unintentionally paves the way for releasing the generator as a Pypi package so people can easily install it, but let's leave that out of scope here.

cargo/poetry.lock
hfiguiere, line 1: does the lockfile really belong here?

pieterdd: In general I'd say yes. The purpose of a lockfile is to enable repeatable installs, pinning dependency resolutions for predictability until you run `poetry update`. Further details are available [here](https://python-poetry.org/docs/basic-usage/#committing-your-poetrylock-file-to-version-control).

pieterdd: Would you prefer I remove the lockfile? My main goal was simplifying the logistics of running `flatpak-cargo-generator.py`, so pinning the dependency resolutions is secondary.

hfiguiere: > Would you prefer I remove the lockfile? 

Yes.

refi64: Fwiw we have the lockfile checked in for node-generator (but that does have quite a few development dependencies).

pieterdd: I'll leave out the lockfile in the interest of moving along the review, but that's a good point.

PR #423, opened 2024-09-18
Fix to handling of packages with caps

jonathon-love:
Currently flatpak builder doesn't handle packages with caps in them properly. i.e. it does not generate the sources appropriately for, say, `PyYAML`.

### DIFF
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -414,10 +414,11 @@ for package in packages:
     is_vcs = True if package.vcs else False
     package_sources = []
     for dependency in dependencies:
-        if dependency in sources:
-            source = sources[dependency]
-        elif dependency.replace('_', '-') in sources:
-            source = sources[dependency.replace('_', '-')]
+        casefolded = dependency.casefold()
+        if casefolded in sources:
+            source = sources[casefolded]
+        elif casefolded.replace('_', '-') in sources:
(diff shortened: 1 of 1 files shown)


dreua:
In which way did it generate the sources wrong?
In the case of pikepdf this change results in `Deprecated` (package name with capital D) vanishing from the generated sources. I now have to figure out if that should never have been there in the first place or if it is now missing due to the change.
https://github.com/dreua/flatpak-module-qpdf-pikepdf/pull/84/files

dreua:
I think this is a regression, pikepdf now fails to build because `Deprecated` is missing in the newly generated dependencies.
Are you sure the keys in `sources` are casefolded? I guess they aren't but haven't run the debugger on it.

bbhtt:
Try https://github.com/flatpak/flatpak-builder-tools/pull/429

dreua:
I just came up with the very small change in #430 which works for me. Now I see #429 which looks a lot more complex. Not sure what issues it needs to solve though, maybe that complexity is necessary.

dreua:
There is a difference in the x-checker-data, but the data checker doesn't seem to care:

```diff
         x-checker-data:
-          name: Deprecated [my version]
+          name: deprecated [@bbhtt's version]
```

bbhtt:
because you are lowercasing it in only one place, anyway that doesn't cause any functional difference.

Issue #428, opened 2024-10-09
Generating dependencies for SecretStorage misses SecretStorage


hadess:
### flatpak-builder version

1.4.4

### Linux distribution and version

Fedora 40

### Affected flatpak-builder tool

pip/flatpak-pip-generator

### Description

```
flatpak-builder-tools/pip/flatpak-pip-generator --yaml "SecretStorage"
```

The generated file is missing `SecretStorage-3.3.3-py3-none-any.whl`.

This is a regression caused by 9a48b5e30a53715f1e71a5b804ff99fa46c430a3 (CC @jonathon-love)

jonathon-love:
hi,

i'm actually not able to reproduce this. this is what i get:

```
# Generated with flatpak-pip-generator --yaml SecretStorage
name: python3-SecretStorage
buildsystem: simple
build-commands:
  - pip3 install --verbose --exists-action=i --no-index --find-links="file://${PWD}"
    --prefix=${FLATPAK_DEST} "SecretStorage" --no-build-isolation
sources:
  - type: file
    url: https://files.pythonhosted.org/packages/54/24/b4293291fa1dd830f353d2cb163295742fa87f179fcc8a20a306a81978b7/SecretStorage-3.3.3-py3-none-any.whl
    sha256: f356e6628222568e3af06f2eba8df495efa13b3b63081dafd4f7d9a7b7bc9f99
  - type: file
    url: https://files.pythonhosted.org/packages/fc/97/c783634659c2920c3fc70419e3af40972dbaf758daa229a7d6ea6135c90d/cffi-1.17.1.tar.gz
    sha256: 1c39c6016c32bc48dd54561950ebd6836e1670f2ae46128f67cf49e789c52824
  - type: file
    url: https://files.pythonhosted.org/packages/de/ba/0664727028b37e249e73879348cc46d45c5c1a2a2e81e8166462953c5755/cryptography-43.0.1.tar.gz
    sha256: 203e92a75716d8cfb491dc47c79e17d0d9207ccffcbcb35f598fbe463ae3444d
  - type: file
    url: https://files.pythonhosted.org/packages/ae/72/2a1e2290f1ab1e06f71f3d0f1646c9e4634e70e1d37491535e19266e8dc9/jeepney-0.8.0-py3-none-any.whl
    sha256: c0a454ad016ca575060802ee4d590dd912e35c122fa04e70306de3d076cce755
  - type: file
[...]

PR #429, opened 2024-10-29
pip: Ignore case and assume lowercase name when package is from pypi

bbhtt:
pypi enforces case-insensitive naming https://peps.python.org/pep-0508/#names however we cannot say the same for vcs packages.

Fixes https://github.com/flatpak/flatpak-builder-tools/issues/428


bbhtt:
this should make both input combinations work:

```
flatpak-pip-generator PyYAML
flatpak-pip-generator pyyaml

flatpak-pip-generator SecretStorage
flatpak-pip-generator secretstorage
```

dreua (approved):
works for me

razzeee:
Seems to also have fixed this for me. Couldn't build before changing to this branch, due to a dependency with an Uppercase name.

hfiguiere:
LGTM

PR #430, opened 2024-10-29
pip: Fix regression with uppercase dependencies

dreua:
Fix regression introduced in https://github.com/flatpak/flatpak-builder-tools/pull/423


bbhtt:
you cannot assume lowercase for everything, only PyPi enforces that, VCS packages might not be case sensitive. See https://github.com/flatpak/flatpak-builder-tools/pull/429

dreua:
That assumption has been made in #423, not by me. But you might be right, there might be someone out there relying on different casing to distinguish their packages. I don't care how this is fixed, as long as it is fixed rather sooner than later.

bbhtt:
yea #423 is wrong, you will need to modify `sources` first, then look for casefolded names, otherwise it matches nothing and will fail for even from PyPi.

PR #431, opened 2024-11-05
Use lower case for source name of package

cedk:
Since 9a48b5e3 the dependencies are tested with .casefold()


hfiguiere:
I merged #429. So this might be irrelevant now.

hfiguiere:
Feel free to rebase it the changes are still relevant after #429

PR #435, opened 2024-11-28
Add none-any.whl

Murmele:
If the file is of type none-any.whl it can be used as well, because it is suitable for all architectures
Files: flatpak-pip-generator

### DIFF
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -83,7 +83,7 @@ def get_tar_package_url_pypi(name: str, version: str) -> str:
     url = 'https://pypi.org/pypi/{}/{}/json'.format(name, version)
     with urllib.request.urlopen(url) as response:
         body = json.loads(response.read().decode('utf-8'))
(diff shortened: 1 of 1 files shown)


pip/flatpak-pip-generator
```diff
@@ -288,14 +290,18 @@ with tempfile.TemporaryDirectory(prefix=tempdir_prefix) as tempdir:
         if not filename.endswith(('bz2', 'any.whl', 'gz', 'xz', 'zip')):
             version = get_file_version(filename)
             name = get_package_name(filename)
-            url = get_tar_package_url_pypi(name, version)
+            try:
+                url = get_tar_package_url_pypi(name, version)
+                print('Downloading {}'.format(url))
+                download_tar_pypi(url, tempdir)
+            except Exception as err:
+                # Can happen if only a arch dependend whl is available like for wasmtime-27.0.2
```
hfiguiere, line 298:
```suggestion
                # Can happen if only an arch dependent wheel is available like for wasmtime-27.0.2
```

PR #436, opened 2024-12-05
Fix some typos in the cargo flatpak manifest

0xfalafel:
Hi !

This commit fixes 2 things in the flatpak manifest:

* In the `build-commands`, cargo build the app with `--release`. But the folder referenced is `/debug/` instead of `/release/`
* In the file `com.flatpak.quickstart.json`, there is a small typo: `qucikstart` instead of `quickstart`


hfiguiere:
good catch!

dreua:
From my point of view it would be great if all commit messages could have a tag or mention of the environment / language affected. This would make it easier to decide whether a commit is relevant for my project or not. Thanks!

PR #440, opened 2025-02-17
pip: Add changelog links from pypi to checker data

dreua:
Working on a feature for data checker to make changes easier to review by adding links to changelogs: https://github.com/flathub-infra/flatpak-external-data-checker/issues/339

My proof of concept works fine already and it obviously makes sense to pull the changelog links from pypi. Let me know what you think, PR for the checker will follow.

### DIFF
--- a/pip/flatpak-pip-generator
+++ b/pip/flatpak-pip-generator
@@ -78,6 +78,24 @@ def get_pypi_url(name: str, filename: str) -> str:
                     return source['url']
         raise Exception('Failed to extract url from {}'.format(url))
 
+def get_pypi_changelog(name: str) -> str:
+    url = 'https://pypi.org/pypi/{}/json'.format(name)
+    with urllib.request.urlopen(url) as response:
+        body = json.loads(response.read().decode('utf-8'))
+        urls = {k.casefold(): v for k, v in body['info']['project_urls'].items()}
+        # Keys as recognized by pypi: 
+        # https://github.com/pypi/warehouse/blob/main/warehouse/templates/packaging/detail.html#L24
+        search_keys = {"changelog", "change log", "changes", "release notes", "news", "what's new", "history"}
+        keys = search_keys.intersection(urls.keys())
+        if len(keys) == 1:
+            return urls[keys.pop()]
+        elif len(keys) > 1:
+            print(f"Warning: Got multiple potential keys for the changelog, picking one at random from {keys}.")
+            return urls[keys.pop()]
+        else:
+            print(f"No changelog url found for '{name}'.")
+
+
 
 def get_tar_package_url_pypi(name: str, version: str) -> str:
     url = 'https://pypi.org/pypi/{}/{}/json'.format(name, version)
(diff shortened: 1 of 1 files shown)


pip/flatpak-pip-generator
```diff
@@ -78,6 +78,24 @@ def get_pypi_url(name: str, filename: str) -> str:
...
+        urls = {k.casefold(): v for k, v in body['info']['project_urls'].items()}
+        # Keys as recognized by pypi: 
+        # https://github.com/pypi/warehouse/blob/main/warehouse/templates/packaging/detail.html#L24
+        search_keys = {"changelog", "change log", "changes", "release notes", "news", "what's new", "history"}
+        keys = search_keys.intersection(urls.keys())
+        if len(keys) == 1:
+            return urls[keys.pop()]
+        elif len(keys) > 1:
+            print(f"Warning: Got multiple potential keys for the changelog, picking one at random from {keys}.")
+            return urls[keys.pop()]
```
bbhtt, line 94: I'd just combine these two cases and pick the first one always. If it's wrong it needs to be manually edited anyways, the warning doesn't particularly mean that it was wrong either.

bbhtt:
Is this ready or not? If so you have to rebase and undraft. I think adding extra metadata to JSON is fine.

pip/flatpak-pip-generator
```diff
@@ -78,6 +78,24 @@ def get_pypi_url(name: str, filename: str) -> str:
                     return source['url']
         raise Exception('Failed to extract url from {}'.format(url))
 
+def get_pypi_changelog(name: str) -> str:
```
bbhtt, line 81: return type is wrong.

pip/flatpak-pip-generator
```diff
@@ -78,6 +78,24 @@ def get_pypi_url(name: str, filename: str) -> str:
...
+    url = 'https://pypi.org/pypi/{}/json'.format(name)
+    with urllib.request.urlopen(url) as response:
+        body = json.loads(response.read().decode('utf-8'))
+        urls = {k.casefold(): v for k, v in body['info']['project_urls'].items()}
+        # Keys as recognized by pypi: 
+        # https://github.com/pypi/warehouse/blob/main/warehouse/templates/packaging/detail.html#L24
```
bbhtt, line 87: are these officially documented anywhere?

bbhtt: ah wait this is the official one.

dreua:
Thanks for the review and good comments, I'll implement them asap! It is not ready because I'm not sure about the naming of the key. In the beginning I assumed that it should be a template (to add the current version), therefore `changelog-url-template`, but I ended up not needing nor implementing any templating here. Therefore I'm leaning toward shortening the key to `changelog-url`. What do you think?

pip/flatpak-pip-generator
dreua: I don't have any data on how many projects might have values for more than one of those keys or if the keys may actually have different semantic meanings. (In some projects on Github I definitely saw differences in changlog and release notes.)
Therefore my intention was to give the users a hint that they might not get the best result or then one they are expecting. It could be rewritten to be more informational than a warning, maybe also the code cleaned up a bit. (Maybe printing all keys and values to be more useful and allow for easy manual fixing afterwards.)

But if you think it is really unnecessary and want the code as short as possible I'm happy to remove it.

bbhtt:
I don't have an opinion about the key name, either works I think. It's going to be generated anyways.

PR #442, opened 2025-02-24
pip: don't consider system_packages for python2

dcbaker:
The freedesktop runtime only provides python3, so this will incorrectly skip important packages when run in python2 mode that are actually needed.


bbhtt:
Curious where are you still using Python 2 with this.

dcbaker:
@bbhtt: building flatpaks out of renpy projects: https://github.com/dcbaker/flatpaker

PR #446, opened 2025-03-10
quote PWD so it can handle spaces

### LINKED ISSUE #445: node script fails to run if PWD contains spaces
It appears as though this line
https://github.com/flatpak/flatpak-builder-tools/blob/aac65cf44cd4e008594a9d9ac1db08e2025067a6/node/flatpak_node_generator/providers/npm.py#L511

is causing a "No such file or directory" error when the PWD contains a space in the path (such as when one of the parent folder names contains a space).

Solution: quote the PWD variable so that its entire value gets stored in the `FLATPAK_BUILDER_BUILDDIR` variable, not just the part up to the first space.

MoralCode:
patch to (hopefully) fix #445


MoralCode:
ok it looks like the combination of these two things fixed the issue

bbhtt:
Squashed the commit and changed the commit message slightly.

PR #463, opened 2025-05-10
deno: add docs to get 100 score in jsr

sigmaSd:


sigmaSd:
Its ok If I push changes directly or should I open a pr each time ?

bbhtt:
It has to go through a PR, pushing directly is blocked via branch protection. If this had only changes in `deno/**` you will be able to merge it on your own.

bbhtt:
Also make sure to prefix the commits with `deno:` I will rewrite it for now

PR #465, opened 2025-05-10
deno: handle peer deps with multiple peers

sigmaSd:


sigmaSd:
@bbhtt can you please add me to the members of this scope https://jsr.io/@flatpak-contrib/~/members currently packaging fails because I'm not

PR #471, opened 2025-05-15
Setup cargo config in cargo generator README

swsnr:
Without this step the configuration file is at the wrong place, and cargo won't use the vendored sources.


bbhtt:
This isn't needed if you set:

```
    build-options:
      env:
        CARGO_HOME: /run/build/$module_name/cargo
```

as written below. But also meh as that's deprecated...

PR #482, opened 2025-08-16
flatpak-cargo-builder: unwrap tomldoc

theoden8:
this was generating invalid hashes because tomlkit.load and tomlkit.parse return TomlDocument instead of dict.


bbhtt:
Where was it producing wrong hashes?

bbhtt:
also prefix the commit message with `cargo: `

bbhtt:
Ah right, I forgot about the YAML output.

PR #490, opened 2025-11-19
pip: topologically sort generated modules based on dependencies

razzeee:


razzeee:
In that example, `python3-passlib` gets put behind `python3-paramiko` 

If I move it to the location on the left, it will work

image


Otherwise it fails with:
```
Requirement already satisfied: markupsafe>=2.1.0 in /usr/lib/python3.12/site-packages (from Flask-Security-Too==5.6.*) (3.0.3)
INFO: pip is looking at multiple versions of flask-security-too to determine which version is compatible with other requirements. This could take a while.
ERROR: Could not find a version that satisfies the requirement passlib>=1.7.4; python_version < "3.13" (from flask-security-too) (from versions: none)
ERROR: No matching distribution found for passlib>=1.7.4; python_version < "3.13"
```

razzeee:
Which points to https://github.com/pallets-eco/flask-security/blob/5.6.2/pyproject-too.toml#L46 as the reason. I guess being able to just call the app, with the targeted python version would be better, but that has it's own problems.

PR #491, opened 2025-11-22
node: Added -S option to set the split size

hfiguiere:
This is useful outside of github, or if the limit ever change.

(for example I'm working with a setup that limit files to 500KB)

Default is unchanged (ie if you don't pass -S)


node/flatpak_node_generator/main.py
bbhtt:
```suggestion
            if fp.tell() >= gen.split_size:
```

hfiguiere: done

PR #493, opened 2025-12-14
nuget: add flag to inject only-arches data into generated files

MoralCode:
This PR is the result of some changes I made locally to the nuget generator to allow me to inject `only-arches` values ([see the module sources docs](https://docs.flatpak.org/en/latest/module-sources.html)) into the generated output.

The reason for this is because I am consistently encountering issues where generating the module sources files for arm machines while using my x86 machine. I have yet to figure out exactly why ~2 dependencies are getting skipped when done from arm (even with the `linux-arm64` runtime configured using existing args). The easiest solution so far is to just regenerate the file on an arm machine and use this new flag to essentially conditionally load it into the main app manifest.

### DIFF
--- a/dotnet/flatpak-dotnet-generator.py
+++ b/dotnet/flatpak-dotnet-generator.py
@@ -46,7 +46,7 @@ def main() -> None:
         default="nuget-sources",
     )
     parser.add_argument(
-        "--only-arches", help="limit the outout to this flatpak arch", default=None
+        "--only-arches", help="Limit the source to this Flatpak arch", default=None
     )
     parser.add_argument(
         "--dotnet-args",


dotnet/flatpak-dotnet-generator.py
```diff
@@ -45,6 +45,9 @@ def main() -> None:
         help="The directory the generated sources file will save sources to",
         default="nuget-sources",
     )
+    parser.add_argument(
+        "--only-arches", help="limit the outout to this flatpak arch", default=None
```
bbhtt:
```suggestion
        "--only-arches", help="Limit the source to this Flatpak arch", default=None
```

MoralCode: I believe there is a list, if unwritten. Heres a document outlining the concept. https://learn.microsoft.com/en-us/dotnet/core/rid-catalog

The list is likely pretty long since the identifers seem to essentially be combinations of OSses and architectures, i.e. `linux-x64`, `unix-x64`, as well as just OSses (`linux`). Same goes for macos and windows too.

Should I only limit it to the linux ones? maybe including Unix too? or should this include macos and windows too?

PR #495, opened 2025-12-29
node: Fix generator ripgrep binary mappings and add missing arches

LorbusChris:
The @vscode/ripgrep package postinstall script expects specific binary variants for each architecture. This patch:

1. Fixes aarch64 to use musl variant instead of gnu
   - Both variants exist, but postinstall expects musl
   - Aligns with x86_64 and i386 (which also use musl)

2. Adds missing architecture mappings for ppc64, riscv64, and s390x
   - These architectures are supported by @vscode/ripgrep
   - Uses gnu variants (only available option for these arches)

Current code causes the postinstall script to fail to find the cached binary on aarch64 and attempt to download from GitHub, which fails in offline build environments (RPM builds, air-gapped systems, etc).

The mappings now match exactly what the postinstall script expects: https://github.com/microsoft/vscode-ripgrep/blob/main/lib/postinstall.js#L38-L47

Assisted-by: Claude Code

### DIFF
--- a/node/flatpak_node_generator/providers/special.py
+++ b/node/flatpak_node_generator/providers/special.py
@@ -303,11 +303,22 @@ class SpecialSourceProvider:
             return match.group(1)
 
         tag = await get_ripgrep_tag(package.version)
+
+        # vscode-ripgrep switched from -gnu to -musl for aarch64 in v1.13.0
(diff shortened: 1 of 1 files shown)


LorbusChris:
@bbhtt mind having another look?

Issue #492, opened 2025-11-24
node generator produce incomplete manifest


hfiguiere:
### flatpak-builder version

1.4.4

### Linux distribution and version

Fedora 42

### Affected flatpak-builder tool

node/flatpak-node-generator.py

### Flatpak-builder manifest URL

https://github.com/hfiguiere/flatpak-builder-node-reproducer.git

### Description

Checkout the repository above. It's a reproducer case, not meant to be a meaningfull flatpak.

Try an initial build.

```
$ ./build.sh
Downloading sources
[...]
Starting build of com.example.FlatpakBuilderNodeReproducer
Cache miss, checking out last cache hit
========================================================================
Building module sass in /home/hub/source/flatpak-builder-node-reproducer/.flatpak-builder/build/sass-6
========================================================================
Running: npm install -g sass
npm error code ENOTCACHED
npm error request to https://registry.npmjs.org/sass failed: cache mode is 'only-if-cached' but no cached response is available.
npm error A complete log of this run can be found in: /run/build/sass/flatpak-node/npm-cache/_logs/2025-11-24T15_35_15_031Z-debug-0.log
Error: module sass: Child process exited with code 1
```

Enter in a build shell

```
$ ./build.sh --build-shell=sass
```

List the cache
```
$ npm cache ls
make-fetch-happen:request-cache:https://registry.npmjs.org/@parcel/watcher-android-arm64/-/watcher-android-arm64-2.5.1.tgz
make-fetch-happen:request-cache:https://registry.npmjs.org/@parcel/watcher-darwin-arm64/-/watcher-darwin-arm64-2.5.1.tgz
make-fetch-happen:request-cache:https://registry.npmjs.org/@parcel/watcher-darwin-x64/-/watcher-darwin-x64-2.5.1.tgz
make-fetch-happen:request-cache:https://registry.npmjs.org/@parcel/watcher-freebsd-x64/-/watcher-freebsd-x64-2.5.1.tgz
make-fetch-happen:request-cache:https://registry.npmjs.org/@parcel/watcher-linux-arm-glibc/-/watcher-linux-arm-glibc-2.5.1.tgz
make-fetch-happen:request-cache:https://registry.npmjs.org/@parcel/watcher-linux-arm-musl/-/watcher-linux-arm-musl-2.5.1.tgz
[...]

Issue #439, opened 2025-02-14
ENOTCACHED error even though the package is cached


suntorytimed:
### flatpak-builder version

1.4.4

### Linux distribution and version

openSUSE Tumbleweed

### Affected flatpak-builder tool

npm/flatpak-npm-generator.py

### flatpak-builder tool cli args

--user --install --force-clean

### Source repository URL

https://github.com/camunda/camunda-modeler

### Flatpak-builder manifest URL

https://github.com/suntorytimed/flathub/tree/io.github.camunda.camunda-modeler

### Description

Even though `webpack-sources` is part of the `generated-sources.json` and is in the cache the `npm install --offline` still fails with:
```
npm error code ENOTCACHED
npm error request to https://registry.npmjs.org/webpack-sources failed: cache mode is 'only-if-cached' but no cached response is available.
```

Fredolx:
I'm having the same issue.

```
Running: npm install --offline --verbose
npm verbose cli /usr/lib/sdk/node20/bin/node /usr/lib/sdk/node20/bin/npm
npm info using npm@10.8.2
npm info using node@v20.19.4
npm verbose title npm install
npm verbose argv "install" "--offline" "--loglevel" "verbose"
npm verbose logfile logs-max:10 dir:/run/build/opentv/flatpak-node/npm-cache/_logs/2025-07-20T00_06_10_278Z-
npm verbose logfile /run/build/opentv/flatpak-node/npm-cache/_logs/2025-07-20T00_06_10_278Z-debug-0.log
npm http fetch GET https://registry.npmjs.org/npm attempt 1 failed with EAI_AGAIN
npm verbose stack Error: request to https://registry.npmjs.org/@tauri-apps%2fplugin-clipboard-manager failed: cache mode is 'only-if-cached' but no cached response is available.
npm verbose stack     at cacheFetch (/usr/lib/sdk/node20/lib/node_modules/npm/node_modules/make-fetch-happen/lib/cache/index.js:12:13)
npm verbose stack     at async fetch (/usr/lib/sdk/node20/lib/node_modules/npm/node_modules/make-fetch-happen/lib/fetch.js:98:7)
npm verbose stack     at async RegistryFetcher.packument (/usr/lib/sdk/node20/lib/node_modules/npm/node_modules/pacote/lib/registry.js:90:19)
[...]

PR #497, opened 2026-01-04
node: document lockfile presence requirement

strugee:
Possibly related: #439, #492

### DIFF
--- a/node/README.md
+++ b/node/README.md
@@ -114,6 +114,14 @@ If you're using npm, you must run this script when the `node_modules` directory 
 If you generate the `generated-sources.json` in CI, you can do this by passing `--package-lock-only`
 to `npm install`.
 
+### Lockfile requirements
+
+If using npm, your lockfile must be **committed** to the project's version control. It's not enough
+to generate it locally and feed it to `flatpak-node-generator` because without it, `npm` will make
+API requests that won't be cached by the generator.
+
+You also must be using at least `package-lock.json` version 2.
+
 ### Caching
 
 flatpak-node-generator will cache many API responses and archives from the server to speed up


strugee:
That's why I said _possibly_ related. Sorry, I could've been way clearer with this description - that's my bad!

I wrote this based on local testing with each lockfile version. It makes sense once you read the npm docs: https://docs.npmjs.com/cli/v11/configuring-npm/package-lock-json#handling-old-lockfiles, https://docs.npmjs.com/cli/v11/configuring-npm/package-lock-json#lockfileversion

strugee:
> Note that the file format changed significantly in npm v7 to track information that would have otherwise required looking in node_modules or the npm registry.

This is the problem. If the lockfile isn't committed to git then npm in fresh clones created by `flatpak-builder` won't have access to this information, and so will trigger registry API fetches. (AFAICT, I haven't read the source code.)

I suppose you could also maintain a `package-lock.json` in the downstream packaging repo and populate it into the build environment somehow. Or vendor `node_modules` into git upstream (but that's bad for other reasons).

I can adjust the language to emphasize committing less if you'd like? But I focused on that since the rest of the docs have some notes on how to generate `package-lock.json` efficiently - for me personally, this caused be to get a bit mixed up and take longer to debug because my brain considered the lockfile to be "present" in the upstream project. Except that with fresh clones for the build environment, it actually wasn't.

bbhtt:
I am not sure what you are trying to say. flatpak-builder by default creates shallow clones so it already will not have access to some git history.

and it works with the project packaged in a tarball as well which does not have any git history.

hfiguiere:
This is definitely NOT #492. I know I filed it.

strugee:
Again, my apologies for the poor communication, I'm really not sure what's going on on. I think it's probably confusing because I'm struggling to explain what I'm saying, but also because I definitely wrote this patch thinking about _my_ particular usecase (for example, I forgot that modules can have tarball sources as well, because I'm packaging an app in a git repo).

The core technical requirement that I want to document is that `package-lock.json` must be present in the build tree of the module when `npm install --offline` is run, and this file must be lockfile version 2 or higher.

The reason for this is that if either of these conditions are not true, `npm install --offline` will want to make API calls to the npm registry, and those API calls will fail because of the `--offline`. The reason for those API calls, presumably, is explained by [this snippet from the npm documentation](https://docs.npmjs.com/cli/v11/configuring-npm/package-lock-json#handling-old-lockfiles):

> When npm detects a lockfile from npm v6 or before during the package installation process, it is automatically updated to fetch missing information from either the `node_modules` tree or (in the case of empty `node_modules` trees or very old lockfile formats) the npm registry.

Because in the Flatpak context you're running a clean build, the `node_modules` tree is empty, so npm has to fetch information from the registry. It is these API calls that are not cached by `flatpak-node-generator` and thus cause `npm install --offline` to fail.

strugee:
To clarify: it does not matter if there is git history available for this file. It does not have to be committed even if you're using a module source of type `git`.

The reason the patch here focuses on committing is because it was poorly written by me, because I was packaging an app with Node.js dependencies, got confused, and then only considered _my personal usecase_ when writing the doc patch. The reason I got confused is because I had a `~/Development/the-app` clone of upstream, and a `~/Development/the-app-flatpak` Flatpak packaging clone, and I had created `~/Development/the-app/package-lock.json` but not committed it. So even though my actual problem was that `~/Development/the-app-flatpak/.flatpak-builder/build/the-app-1/package-lock.json` _didn't_ exist, it took me a while to figure this out because my brain thought it _did_ exist since I had created `~/Development/the-app/package-lock.json`. If I had committed and pushed `~/Development/the-app/package-lock.json` and updated the Flatpak manifest to point at that commit, I would have solved my problem, which is why when I was writing this patch I was super focused on committing the file.

Again, with the benefit of some time and sleep I see how the patch is confusing given tarball sources, etc. - happy to rewrite it to be better, as long as we're on the same page about the actual technical problem.

---

Another clarification: I said that #439 and #492 might _possibly_ be related because the `package-lock.json` issue here caused `ENOTCACHED`, and #439 and #492 mentioned `ENOTCACHED`. I didn't fully look into either of those issues, so I did not claim that this PR was definitely the root cause of those issues, nor did I comment on those issues in order to point people subscribed to them at this PR.

PR #520, opened 2026-03-26
gradle: Don't include missing resource URLs

Newbytee:
If a given URL cannot be downloaded, Gradle will report it as the resource being missing. Respect this when finding URLs so we don't end up with URLs that are 404s.

Also make the URL list a set to facilitate this. It doesn't need to contain duplicates anyway.

### DIFF
--- a/gradle/flatpak-gradle-generator.py
+++ b/gradle/flatpak-gradle-generator.py
@@ -83,23 +83,32 @@ def main():
     for arch in req_flatpak_arches:
         req_gradle_arches.append(flatpak_arch_to_gradle_arch(arch))
 
-    urls = []
+    urls = set()
     urls_arch = {}
+    missing_resources = set()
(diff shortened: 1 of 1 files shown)


hadess:
> If a given URL cannot be downloaded, Gradle will report it as the resource being missing. Respect this when finding URLs so we don't end up with URLs that are 404s.

Do you have a file on which we could test this?

Newbytee:
> > If a given URL cannot be downloaded, Gradle will report it as the resource being missing. Respect this when finding URLs so we don't end up with URLs that are 404s.
> 
> Do you have a file on which we could test this?

You mean like a log? I'm pretty sure this was the codebase I was working with at the time: https://github.com/zacharee/SamloaderKotlin

hadess:
> You mean like a log? I'm pretty sure this was the codebase I was working with at the time: https://github.com/zacharee/SamloaderKotlin

Yes, I'd need a log which I could parse and compare before/after this patch to see whether it's working correctly.

PR #528, opened 2026-04-25
pip: Fix typos

a-andre:


bbhtt:
I noticed you were using it for gajim, I added handling for the platform_system marker: https://github.com/flatpak/flatpak-builder-tools/commit/1da0b3c8146fb26c071a358e567d2175af8dec5c, and the missing handling for pyproject.toml branch: https://github.com/flatpak/flatpak-builder-tools/commit/50a07682428880637a315e4225499b719317344f.

a-andre:
@bbhtt thank you.

Any plans to make the script installable via pip, so we don't have to maintain a copy of it in Gajim's repository? At the moment, the wheel file created from `pyproject.toml` doesn't include the script.

bbhtt:
It's not supposed to be installable script and there aren't any plans to do that.

You can curl it and run via pip or use `uv`. It has inlined dependencies in the top so uv (and others) will automatically handle them.

PR #537, opened 2026-05-12
feat/node: pnpm store v11 support

### LINKED ISSUE #535: Support for pnpm store v11
Since pnpm v11, it has moved to a [new store system](https://pnpm.io/blog/releases/11.0#store-v11), which is not supported by flatpak-node-generator now.

YUCLing:
I've only tested with my project's lockfile, works as intended.

close #535


node/flatpak_node_generator/providers/pnpm.py
```diff
@@ -93,6 +95,16 @@ def process_lockfile(self, lockfile_path: Path) -> Iterator[Package]:
...
+        supported_store_versions = _STORE_VERSION_BY_LOCKFILE[major]
+        if self.store_version is None:
+            self.store_version = supported_store_versions[0]
+        elif self.store_version not in supported_store_versions:
+            supported = ', '.join(str(v) for v in sorted(supported_store_versions))
+            raise ValueError(
+                f"{lockfile_path}: lockfileVersion {raw_version} doesn't support store version {self.store_version}. "
+                f'Supported versions: {supported}.'
+            )
+
```
bbhtt, line 110: a better design is to make this _resolve_store_version and do the resolution in create_module_provider(), process_lockfile can return Package object with the resolved lockfile version

the asserts below are a bit silly and process_lockfile is both mutating and validating.

something like this:

```python
    def create_module_provider(
        self, gen: ManifestGenerator, special: SpecialSourceProvider
    ) -> PnpmModuleProvider:
    lockfile_version = self.lockfile_provider.lockfile_version
    store_version = _resolve_store_version(
        lockfile_version, self.options.lockfile.store_version
    )
    return PnpmModuleProvider(gen, special, self.lockfile_root, store_version)
```

node/flatpak_node_generator/populate_pnpm_store.py
```diff
@@ -104,47 +264,107 @@ def _process_tarball(
...
-            url_idx_path = os.path.join(url_idx_dir, 'integrity.json')
-            with open(url_idx_path, 'w', encoding='utf-8') as out:
-                json.dump(index_data, out)
+    if store_version == 'v11':
+        assert index_db is not None
+        # pnpm v11 store index keys use the raw package name (e.g. @scope/pkg),
+        # not the filesystem-sanitized form (@scope+pkg), since keys are stored
+        # in SQLite, not as filenames.
+        raw_pkg_id = f'{pkg_name}@{pkg_version}'
+        key = f'{integrity}\t{raw_pkg_id}'
```
bbhtt: none integrity would be a bug here.

node/flatpak_node_generator/main.py
```diff
@@ -113,6 +113,10 @@ async def _async_main() -> None:
         help='Use the ChromeDriver version associated with the given '
         'Electron version for node-chromedriver',
     )
+    parser.add_argument(
+        '--pnpm-store-version',
+        help='Specify which store version to use for pnpm. Only supports lockfile v9, defaults to v10.',
```
bbhtt: "Specify the store version for pnpm v9 lockfile." and then (default: ) import something from pnpm to specify the default.

node/pyproject.toml
```diff
@@ -12,6 +12,7 @@ requires-python = "<4.0,>=3.10"
 dependencies = [
     "aiohttp<4.0.0,>=3.9.0",
     "pyyaml<7.0,>=6.0",
+    "msgpack (>=1.1.2,<2.0.0)",
```
bbhtt: this is unused outside tests right nowm, it should be in dev dependencies and, `msgpack  = ">=1.1.2,<2"`

node/flatpak_node_generator/populate_pnpm_store.py
```diff
@@ -6,24 +6,175 @@
...
+    is enabled (pnpm's config).  Standard msgpack uses 0x00-0x7F for positive
+    fixints, but this packer caps them at 0x3F and emits a uint 8 prefix (0xcc)
+    for 64-255 to avoid aliasing with record ID bytes.
+    """
+
+    def _pack_int(val: int) -> bytes:
+        # Positive fixint capped at 0x3F — 0x40-0x7F are msgpackr record IDs
+        if 0 <= val <= 0x3F:
+            return val.to_bytes(1, 'big')
+        if -32 <= val < 0:
```
bbhtt, line 33: should change these to elifs for the whole function.

node/flatpak_node_generator/populate_pnpm_store.py
```diff
@@ -6,24 +6,175 @@
...
+
+    The ``files`` field must be a Map so pnpm can iterate it with ``for..of``.
+    Everything else must use record extensions so dot-notation works
+    (e.g. ``pkgIndex.algo``, ``info.digest``, ``manifest.name``).
+    """
+
+    def _record(obj: Mapping[str, object], struct_id: int) -> bytes:
+        """Pack a dict as a msgpackr record extension (plain object in JS)."""
+        keys = list(obj.keys())
+        result = b'\xd4\x72' + struct_id.to_bytes(1, 'big')
```
bbhtt: can you use a named constant for this?

node/flatpak_node_generator/populate_pnpm_store.py
```diff
@@ -6,24 +6,175 @@
...
+      (iterable, supports ``for..of``)
+    - record extensions (0x72) as plain objects (dot-notation access)
+
+    The ``files`` field must be a Map so pnpm can iterate it with ``for..of``.
+    Everything else must use record extensions so dot-notation works
+    (e.g. ``pkgIndex.algo``, ``info.digest``, ``manifest.name``).
+    """
+
+    def _record(obj: Mapping[str, object], struct_id: int) -> bytes:
+        """Pack a dict as a msgpackr record extension (plain object in JS)."""
```
bbhtt:
```suggestion
```

node/tests/test_populate_pnpm_store.py
```diff
@@ -1,10 +1,192 @@
...
+import msgpack  # type: ignore[import-untyped]
+
+from flatpak_node_generator.populate_pnpm_store import (
+    _msgpack_pack,
+    _pack_v11_store_entry,
+    _process_tarball,
+)
+
+
+def _decode_v11(data: bytes) -> dict[str, object]:
```
bbhtt, line 17: can't you only manually decode the record extension wrapper and use msgpack directly for the rest?

node/tests/test_populate_pnpm_store.py
```diff
@@ -179,3 +361,260 @@ def test_process_tarball_with_long_path(tmp_path: Path) -> None:
...
+
+    store_dir.mkdir(parents=True, exist_ok=True)
+    db = sqlite3.connect(str(store_dir / 'index.db'))
+    db.execute('PRAGMA busy_timeout=5000')
+    db.execute('PRAGMA journal_mode=WAL')
+    db.execute('PRAGMA synchronous=NORMAL')
+    db.execute('PRAGMA mmap_size=536870912')
+    db.execute('PRAGMA cache_size=-32000')
+    db.execute('PRAGMA temp_store=MEMORY')
+    db.execute('PRAGMA wal_autocheckpoint=10000')
```
bbhtt: why are these needed? mmap_size, cache_size and wal_autocheckpoint?

bbhtt:
Adding some integration tests using v11 would be nice.

Also ccing @nukeop @andersk if they want to test / review since you tested last time.

node/flatpak_node_generator/populate_pnpm_store.py
```diff
@@ -6,24 +6,175 @@
...
 import sys
 import tarfile
 import time
+from collections.abc import Mapping
 
 _SANITIZE_RE = re.compile(r'[\\/:*?"<>|]')
 _MAX_LENGTH_WITHOUT_HASH = 120
 
 
+def _msgpack_pack(obj: object) -> bytes:
```
bbhtt, line 20: why can't this use `msgpack.packb()`? only deviation seems the fixint cap isn't it?

YUCLing: Is it okay to introduce a dependency in this script? I'm not sure about that

bbhtt: why should it be a problem? the entire thing is installed together so as long as the dependency is satisfied by the toplevel pyproject.toml it should just work when launching.

node/flatpak_node_generator/populate_pnpm_store.py
YUCLing: Ahh, actually it can't be nullable, made a mistake here

node/flatpak_node_generator/populate_pnpm_store.py
bbhtt: ah nevermind, i forgot that it is executed during build, then indeed you can't add a dependency.

YUCLing: > only deviation seems the fixint cap isn't it?

Right, about this, msgpackr from Node.JS doesn't seem to be happy if this fixint cap is gone. It will cause `Unexpected end of MessagePack` error in pnpm

node/tests/test_populate_pnpm_store.py
YUCLing: I'll do a double-confirm on these configs from pnpm

node/flatpak_node_generator/populate_pnpm_store.py
bbhtt: actually not being able to use it is more annoying, what I would do is put `store_version = manifest['store_version']` in the top and then

```

if store_version == 'v11':
    try:
        import msgpack
    except ImportError:
        print(
            'msgpack is required for pnpm store v11 but is not found.\n'
            'Please include and install it in the manifest',
            file=sys.stderr,
        )
        sys.exit(1)
 ```

 then add in readme showing how it is done.

Comment (YUCLing) [reply node/flatpak_node_generator/populate_pnpm_store.py:20]: I'll try and see if `msgpack` is compatible with `msgpackr` then

**Comment by nukeop** [comment]
I just tested this with Nuclear ([nukeop/nuclear](https://github.com/nukeop/nuclear)), pnpm 11.0.6, lockfile v9.

The store gets populated (index.db is created, pnpm install --offline gets past resolution), but pnpm fails when it tries to extract packages from the store:

```
Running: pnpm install --offline --frozen-lockfile
Scope: all 14 workspace projects
Lockfile is up to date, resolution step is skipped
Packages: +1103
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
[ERR_PNPM_NO_OFFLINE_TARBALL] A package is missing from the store but cannot download it in offline mode. The missing package may be downloaded from https://registry.npmjs.org/@ianvs/prettier-plugin-sort-imports/-/prettier-plugin-sort-imports-4.7.1.tgz.
```

That's the first package alphabetically, so it's not finding any packages at all. The index lookup seems to succeed (it gets past resolution and starts linking), but the files in v11/files/ aren't matching what pnpm expects.

References:

- Full CI run: https://github.com/nukeop/nuclear/actions/runs/25790498744
- Workflow config: https://github.com/nukeop/nuclear/blob/feat/store-v11/.github/workflows/test-flatpak.yml
- Generation command: flatpak-node-generator pnpm --pnpm-store-version v11 -o flatpak/pnpm-sources.json pnpm-lock.yaml
- Flatpak manifest: https://github.com/nukeop/nuclear/blob/feat/store-v11/flatpak/com.nuclearplayer.Nuclear.yml

Let me know if I can help with debugging in any way.

**Comment by YUCLing** [comment]
@nukeop Thanks for testing!

> That's the first package alphabetically, so it's not finding any packages at all.

Not really, pnpm is using workers to look up indexes and read package information, so it depends on which worker fails first.

> The index lookup seems to succeed (it gets past resolution and starts linking), but the files in v11/files/ aren't matching what pnpm expects.

Actually it can also be pnpm fails to find the entry it needs from index, but it doesn't tell you what's exactly wrong. I will look into it to find out what's actually wrong.

**Comment by YUCLing** [comment]
@nukeop Hi there! I just pinned down the issue you encountered! It's not that store is not working, actually, it's working perfectly well!

The problem is that pnpm v11 no longer reads `store-dir` from `.npmrc`, thus it's not using the store at all. You can now workaround by using pnpm's `--store-dir` argument

<details>
<summary>Example</summary>

```
sh-5.2$ pnpm install --offline --frozen-lockfile --store-dir /home/rain/projects/nuclear-music-player-fp/nuclear/flatpak/.flatpak-builder/build/nuclear-music-player-1/flatpak-node/pnpm-store
Scope: all 14 workspace projects
Lockfile is up to date, resolution step is skipped
Packages: +1103
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Progress: resolved 1103, reused 1097, downloaded 0, added 1103, done

devDependencies:
+ @ianvs/prettier-plugin-sort-imports 4.7.1
+ @tailwindcss/vite 4.2.4
+ @vitejs/plugin-react 5.2.0
+ @vitest/coverage-v8 4.1.5
+ eslint 10.3.0
+ husky 9.1.7
+ jsdom 26.1.0
+ lint-staged 16.4.0
+ prettier-plugin-astro 0.14.1
+ prettier-plugin-tailwindcss 0.8.0
+ tailwindcss 4.2.4
+ turbo 2.9.7
+ typescript 5.9.3
+ vite 7.3.3
+ vite-plugin-dts 4.5.4
+ vite-plugin-svgr 4.5.0
+ vitest 4.1.5

. prepare$ husky
└─ Done in 66ms
Done in 3.9s using pnpm v11.0.6
```

Had some free space issues, so most files are moved by pnpm already, that's why it appears to be finished fast

</details>

**Comment by nukeop** [comment]
You were right, with the correction you suggested the build passes just fine.

https://github.com/nukeop/nuclear/actions/runs/25802919961/job/75797639120

File: node/flatpak_node_generator/providers/pnpm.py (hunk 1 above)
Comment (YUCLing) [reply node/flatpak_node_generator/providers/pnpm.py:110]: Does this pattern look good to you? Or I'd better not edit Lockfile class

```python
@dataclass(frozen=True, eq=True)
class Lockfile:
    path: Path
    version: int
    cache_key: str | None = None
    store_version: str | None = None

class PnpmLockfileProvider(LockfileProvider):
    // ...
    def process_lockfile(self, lockfile_path: Path) -> Iterator[Package]:
        // ...
        lockfile = Lockfile(lockfile_path, major, store_version=self.store_version)

class PnpmModuleProvider(ModuleProvider):
    // ...
    async def generate_package(self, package: Package) -> None:
        source = package.source

        if package.lockfile.store_version is None:
            raise TypeError("pnpm expects lockfile provides store version")

        self._store_version = package.lockfile.store_version
```

bbhtt: self._store_version is being set on every package, it should be done only once that's why I was saying to move it to create_module_provider(), the alternative is doing it once in generate_package() like this

```
self._store_version: str | None = None

async def generate_package(self, package: Package) -> None:
    if self._store_version is None:
        sv = package.lockfile.store_version
        if sv is None:
            raise TypeError("pnpm expects lockfile to provide store_version")
        self._store_version = sv

....
```

bbhtt: the mutation is happening inside PnpmModuleProvider instance.

node/flatpak_node_generator/populate_pnpm_store.py
YUCLing: Wanna confirm on this, I think elifs can be more confusing in this case, since every branch is simply a direct `return`.

node/flatpak_node_generator/populate_pnpm_store.py
YUCLing: Sadly, I just confirmed that `msgpack.packb` doesn't support the MessagePack format used by pnpm out of box.

Using `msgpack` will cause this error in pnpm: `[ERR_PNPM_READ_FROM_STORE] files is not iterable`, as it doesn't support the Record type used by `msgpackr` (and there are also other JS types that aren't supported by `msgpack`)

I should simply remove the usage of `msgpack` in tests, since it seems to be testing for nothing.

bbhtt: Ok sounds good.

node/flatpak_node_generator/populate_pnpm_store.py
bbhtt: hmm I thought I saw some where it wasn't returning on every branch, I may have commented on the wrong line. Changing those would be nice.

node/flatpak_node_generator/providers/pnpm.py
bbhtt: this isn't fully resolved, the validation is still being performed on every call instead of per-lockfile once.

node/flatpak_node_generator/populate_pnpm_store.py
```diff
@@ -6,24 +6,174 @@
...
+    for fname, finfo in files.items():
+        files_map_bytes += _msgpack_pack(fname) + _record(finfo, 0x41)
+
+    # Outer store_entry as record (struct_id 0x40)
+    store_entry_keys = ['algo', 'requiresBuild', 'files']
+    if manifest is not None:
+        store_entry_keys.append('manifest')
+
+    result = b'\xd4\x72\x40' + _msgpack_pack(store_entry_keys)
+    result += _msgpack_pack('sha512')  # algo
```
bbhtt: this is hardcoded to sha512 but the key later is `{integrity_algo}-{integrity}`?

node/flatpak_node_generator/providers/pnpm.py
```diff
@@ -239,7 +256,9 @@ def _add_store_population_script(self) -> None:
             entry: dict[str, str] = {
                 'name': info.name,
                 'version': info.version,
-                'integrity_hex': info.integrity.digest,
+                'integrity': info.integrity.to_base64(),
```
bbhtt, line 267: is this rename necessary?

node/tests/test_populate_pnpm_store.py
```diff
@@ -179,3 +395,230 @@ def test_process_tarball_with_long_path(tmp_path: Path) -> None:
...
+    tar_path = tmp_path / 'pkg.tgz'
+    store_dir = tmp_path / 'store' / 'v11'
+    pkg_json = json.dumps({'name': 'real-pkg', 'version': '1.2.3'})
+
+    _create_tarball(
+        tar_path,
+        {'package/package.json': pkg_json, 'package/index.js': "console.log('hello');"},
+    )
+
+    store_dir.mkdir(parents=True, exist_ok=True)
```
bbhtt, line 465: the db is being created here manually instead of exercising the code path that actually runs, it's why i was saying to add an integration test earlier.

node/flatpak_node_generator/providers/pnpm.py
bbhtt: and it is still being mutated

bbhtt: i pushed a commit to fix this. you can squash it with yourself if it works.

node/flatpak_node_generator/providers/pnpm.py
YUCLing: To avoid confusion, since now there are three fields for integrity

node/tests/test_populate_pnpm_store.py
YUCLing: Added `test_populate_store_v11`

node/flatpak_node_generator/providers/pnpm.py
```diff
@@ -263,7 +290,12 @@ def _add_store_population_script(self) -> None:
         )
 
     def _add_pnpm_config(self) -> None:
-        self.gen.add_command(f'echo "store-dir=$PWD/{self.store_dir}" >> .npmrc')
+        if self._store_version == 'v11':
+            self.gen.add_command(
+                f'echo "storeDir=$PWD/{self.store_dir}" >> pnpm-workspace.yaml'
```
andersk, line 295: This is not YAML syntax.


```suggestion
                f'echo "storeDir: $PWD/{self.store_dir}" >> pnpm-workspace.yaml'
```

YUCLing: Ahh, my mistake here, you can submit a PR if you like, I'm not able to make a PR atp

PR #538, opened 2026-05-21
cargo: restructure as package

abn:
Restructuring the cargo generator so that users can use one of the following to install similar to the node generator.

```shell
pipx install git+https://github.com/flatpak/flatpak-builder-tools.git#subdirectory=cargo
uvx --from "git+https://github.com/flatpak/flatpak-builder-tools.git#subdirectory=cargo" flatpak-cargo-generator /path/to/Cargo.lock -o cargo-sources.json
```


bbhtt:
this is going to break current uses of the script. there has been several issues about changing various generators to proper packages and each time I answered no. you can look it up.

the node generator started as a proper package for a long time.

also you can already do:

```
uv run https://raw.githubusercontent.com/flatpak/flatpak-builder-tools/refs/heads/master/cargo/flatpak-cargo-generator.py --help
```

abn:
@bbhtt thanks for the response, apologies for the noise.

PR #553, opened 2026-08-31
node: Add --no-playwright-browsers flag

threema-sebastian:
To avoid fetching the playwright browsers in cases where we don't need them in the final flatpack sandbox.


bbhtt:
can you add a test too?

threema-sebastian:
sorry wasn't able to get back to this sooner, will provide tests next week

